PR #112, opened 2018-04-04
[FIX] TypeError: this.history[keys[n]][indices[n]].dispose is not a f…

rodrigopivi:
Hi, thanks for this awesome project.
I am having a problem, when ported this model to tfs: https://github.com/bhaveshoswal/CNN-text-classification-keras
after the model trains i get the error `TypeError: this.history[keys[n]][indices[n]].dispose is not a function`
Tried using and not using `dispose` and `tidy`, but getting the same error, the only solution i found is what this PR does. Thanks.

---
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs-layers/112)

### DIFF
--- a/src/callbacks.ts
+++ b/src/callbacks.ts
@@ -377,7 +377,9 @@ export class History extends Callback {
     }
     const values = await Promise.all(promises);
     for (let n = 0; n < values.length; ++n) {
-      (this.history[keys[n]][indices[n]] as Tensor).dispose();
+      if ((this.history[keys[n]][indices[n]] as Tensor).dispose) {
+        (this.history[keys[n]][indices[n]] as Tensor).dispose();
+      }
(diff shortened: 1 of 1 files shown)


rodrigopivi:
I signed it!

caisq:
@rodrigopivi Thanks for letting us know of this issue. Can you share us the JavaScript or TypeScript code of the ported model, so we can look at under what circumstances the content of the `History` object isn't a `Tensor`?

rodrigopivi:
@caisq here is an example https://github.com/rodrigopivi/tsjs-cnn-text-classification

`npm start` runs the webpack dev serve on port 3000, the funny thing, is that if i inline the json dataset instead of loading it through `require` i don't get the error. Thanks for your review of this issue.

nsthorat:
Just curious if we resolved this / know what this is from.

PR #131, opened 2018-04-17
Fix a bug in custom-loss support

caisq:
* losses.get() now does the correct thing when a custom loss function is
  passed in

Fixes: https://github.com/tensorflow/tfjs/issues/198

---
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs-layers/131)

### DIFF
--- a/src/engine/training.ts
+++ b/src/engine/training.ts
@@ -325,8 +325,8 @@ export function sliceArraysByIndices(
   } else {
     // TODO(cais): indices should be oa pre-constructed Tensor1D to avoid
     //   tensor1d() calls.
-    return K.gather(arrays,
-        indices.dtype === 'int32' ? indices : indices.toInt());
+    return K.gather(
+        arrays, indices.dtype === 'int32' ? indices : indices.toInt());
   }
 }
 
@@ -595,7 +595,8 @@ export interface ModelCompileConfig {
    * The loss value that will be minimized by the model will then be the sum
    * of all individual losses.
    */
-  loss: string|string[]|{[outputName: string]: string};
+  loss: string|string[]|{[outputName: string]: string}|LossOrMetricFn|
+      LossOrMetricFn[]|{[outputName: string]: LossOrMetricFn};
 
   /**
    * List of metrics to be evaluated by the model during training and testing.
@@ -622,7 +623,8 @@ export interface ModelCompileConfig {
 @doc({heading: 'Models', subheading: 'Classes'})
 export class Model extends Container {
   optimizer: Optimizer;
-  loss: string|string[]|{[outputName: string]: string};
+  loss: string|string[]|{[outputName: string]: string}|LossOrMetricFn|
+      LossOrMetricFn[]|{[outputName: string]: LossOrMetricFn};
   lossFunctions: LossOrMetricFn[];
 
   // TODO(cais): These private variables should probably not have the string
@@ -680,7 +682,8 @@ export class Model extends Container {
 
     // Prepare loss functions.
     let lossFunctions: LossOrMetricFn[] = [];
-    if (!Array.isArray(config.loss) && typeof config.loss !== 'string') {
+    if (!Array.isArray(config.loss) && typeof config.loss !== 'string' &&
+        typeof config.loss !== 'function') {
       config.loss = config.loss as {[outputName: string]: string};
       for (const name in config.loss) {
         if (this.outputNames.indexOf(name) === -1) {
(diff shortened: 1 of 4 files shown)


nsthorat:
High level comment: Losses should return a `Scalar`, LossOrMetricFn now returns a Tensor. We can make this strict if you change that type (and throw compiler error for TS users)

I also think it makes sense to use "docinline" as a jsdoc tag so that we inline that type in the docs.

---

Reviewed 2 of 4 files at r1.
Review status: 2 of 4 files reviewed at latest revision, all discussions resolved.

---



*Comments from [Reviewable](https://beta.reviewable.io/reviews/tensorflow/tfjs-layers/131#-:-LAKRQ4d4bTqlsGcFqPm:bgbjvz4)*

caisq:
I don't think Keras loss functions always return scalars. Take `keras.losses.mean_squared_error` for example: it does reduce mean over only the last axis. So it returns non-scalars for any inputs with rank >= 2:
https://github.com/keras-team/keras/blob/master/keras/losses.py#L14

I believe this is the reason why we typed LossAndMetricFn as returning a Tensor. The logic for further reducing the loss to Scalar is somewhere else (in model's training logic).

---

Review status: 2 of 4 files reviewed at latest revision, all discussions resolved.

---



*Comments from [Reviewable](https://beta.reviewable.io/reviews/tensorflow/tfjs-layers/131#-:-LAKYiAV0mLMQ-Y8aVr5:bgenaxl)*

caisq:
Non-scalar return shape of the `mean_squared_error` loss function can be confirmed with the following code:

```python
import keras
import tensorflow as tf

tf.enable_eager_execution()

x = tf.constant([[1, 2], [3, 4]])
y = tf.constant([[8, 7], [6, 5]])

print(keras.losses.mean_squared_error(x, y))
```

---

Review status: 2 of 4 files reviewed at latest revision, all discussions resolved.

---



*Comments from [Reviewable](https://beta.reviewable.io/reviews/tensorflow/tfjs-layers/131#-:-LAKZMfUEBSBQbRAyt3z:brsbddf)*

nsthorat:
:lgtm_strong:

Got it. Makes sense then. Will you add the docinline jsdoc to the type so it appears in the docs?

---

Review status: 2 of 4 files reviewed at latest revision, all discussions resolved.

---

*[src/engine/training_test.ts, line 17 at r1](https://beta.reviewable.io/reviews/tensorflow/tfjs-layers/131#-LAKZqPIARkazBxcPopC:-LAKZqPIARkazBxcPopD:ba8yke2) ([raw file](https://github.com/tensorflow/tfjs-layers/blob/db8e5a4e57eaf0ae75bc85d34ecbbf844e320717/src/engine/training_test.ts#L17)):*
> ```XML
> // tslint:disable:max-line-length
> import {abs, mean, Scalar, scalar, SGDOptimizer, Tensor, tensor1d, tensor2d, tensor3d, zeros} from '@tensorflow/tfjs-core';
> import {expectArraysClose} from '@tensorflow/tfjs-core/dist/test_util';
> ```

you can import test_util directly (and not reach into dist)

---


*Comments from [Reviewable](https://beta.reviewable.io/reviews/tensorflow/tfjs-layers/131#-:-LAK_-Vu-_JUKOm0W4Dw:bx34vtu)*

caisq:
Inline doc updated.

---

Review status: 2 of 4 files reviewed at latest revision, 1 unresolved discussion, all commit checks successful.

---

*[src/engine/training_test.ts, line 17 at r1](https://beta.reviewable.io/reviews/tensorflow/tfjs-layers/131#-LAKZqPIARkazBxcPopC:-LAO0-4v0VZvpWO5ViET:b-bbqate) ([raw file](https://github.com/tensorflow/tfjs-layers/blob/db8e5a4e57eaf0ae75bc85d34ecbbf844e320717/src/engine/training_test.ts#L17)):*

you can import test_util directly (and not reach into dist)
</blockquote></details>

Done. I've relied on VSCode a little too much recently.

---


*Comments from [Reviewable](https://beta.reviewable.io/reviews/tensorflow/tfjs-layers/131#-:-LAO0kyXCO2ETNOta9nw:b-95iwkj)*

davidsoergel:
:lgtm_strong:

---

Review status: 2 of 4 files reviewed at latest revision, 1 unresolved discussion, some commit checks pending.

---



*Comments from [Reviewable](https://beta.reviewable.io/reviews/tensorflow/tfjs-layers/131#-:-LAO1hcHCqU8Fn-5st6k:b-jbsf8z)*

PR #140, opened 2018-04-20
Introduce Serializable base class

ericdnielsen:
We want to eliminate the use of constructor.name through the layers code.  This is the first step towards that.

---
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs-layers/140)

### DIFF
--- a/src/backend/tfjs_backend.ts
+++ b/src/backend/tfjs_backend.ts
@@ -396,8 +396,8 @@ export function sliceAlongFirstAxis(
           [size, array.shape[1], array.shape[2], array.shape[3]]);
     default:
       throw new ValueError(
-          `sliceAlongFirstAxis() received an unsupported subtype of Tensor: ` +
-          `${array.constructor.name}`);
+          `sliceAlongFirstAxis() received an unsupported tensor rank: ` +
+          `${array.rank}`);
   }
 }
 
@@ -426,8 +426,8 @@ export function sliceAlongLastAxis(
           [array.shape[0], array.shape[1], array.shape[2], size]);
     default:
       throw new ValueError(
-          `sliceAlongLastAxis() received an unsupported subtype of Tensor: ` +
-          `${array.constructor.name}`);
+          `sliceAlongLastAxis() received an unsupported tensor rank: ` +
+          `${array.rank}`);
   }
 }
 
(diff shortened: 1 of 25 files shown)


ericdnielsen:
I have two concerns with this change that I'd like feedback on

1) Its yet another hardcoded string that has to match.  Where we were hoping to eliminate these, instead we're adding more.  Places with deeper inheritance hierarchies have a higher likelihood of people forgetting to override a class name,

2) the fromConfig half of the Serializable contract is not captures.  Its a static method and from my explorations, static methods and JS don't act as intended.  So it was left out for now.

caisq:
:lgtm_strong:

Thanks for doing this! I think the benefit outweighs the burden of maintaining another string constant. Before you submit, can you do a global search for `constructor.name` to make sure we removed all of them in this PR.

---

Review status: 0 of 24 files reviewed at latest revision, all discussions resolved, some commit checks failed.

---

*[src/types.ts, line 330 at r1](https://beta.reviewable.io/reviews/tensorflow/tfjs-layers/140#-LAZVNv198Fox3-L60zE:-LAZVNv198Fox3-L60zF:bx5lq6y) ([raw file](https://github.com/tensorflow/tfjs-layers/blob/952c8210c99e9a5892b440138092a094ab4a1807/src/types.ts#L330)):*
> ```
> abstract getClassName(): string;
> ```

Please add a doc string to describe return type and what it means.

---


*Comments from [Reviewable](https://beta.reviewable.io/reviews/tensorflow/tfjs-layers/140#-:-LAZWllw5NlCWjQ9lH7X:b-pay0x3)*

bileschi:
*[src/types.ts, line 331 at r2](https://beta.reviewable.io/reviews/tensorflow/tfjs-layers/140#-LAZrH6HAZUuRH9_umwx:-LAZrH6HAZUuRH9_umwy:bfvfzxb) ([raw file](https://github.com/tensorflow/tfjs-layers/blob/2a83ee815dc3d9c1648e5b2e2a6d83ee4a3958d6/src/types.ts#L331)):*
> ```
> ot
> ```
nit spelling

---


*Comments from [Reviewable](https://beta.reviewable.io/reviews/tensorflow/tfjs-layers/140)*

bileschi:
This is much better!  Thanks.

---







*Comments from [Reviewable](https://beta.reviewable.io/reviews/tensorflow/tfjs-layers/140#-:-LAZrv5l6RkYYgqFL16P:bjukxxq)*

PR #166, opened 2018-05-08
Backend mem leak test

bileschi:
Adds a test structure at the top of tfjs_backend_test.ts to check for memory leaks. Two tests groups are created, one for functions which are expected to create 0 tensors, and one for functions which are expected to create exactly 1.   For functions which create larger structures of additional tensors, it expected that the caller might test for memory leakage, as is done for `qr` and `normalizeBatchInTraining`.

Some functions need to create static kept tensors on their first execution, thus the pattern of running twice, and only counting the additional load on the second execution.

Memory leaks are fixed for those functions which were leaking.

---
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs-layers/166)

### DIFF
--- a/src/backend/tfjs_backend_test.ts
+++ b/src/backend/tfjs_backend_test.ts
@@ -27,6 +27,105 @@ import * as K from './tfjs_backend';
 
 const CT = ConcreteTensor;
 
+describe('Memory leak check.  Functions with 0 tensors output:', () => {
+  // Reusable arguments
+  const x1D = tensor1d([4, 3, 2, 1]);
+  // Functions that should produce no new Tensors.
+  // tslint:disable-next-line:no-any
+  const zeroOutMap: { [key: string]: () => any } = {};
+  zeroOutMap['countParms'] = () => K.countParams(x1D);
+  zeroOutMap['ndim'] = () => K.ndim(x1D);
+  zeroOutMap['normalizeAxis'] = () => K.normalizeAxis(x1D, 0);
+  zeroOutMap['shape'] = () => K.shape(x1D);
+
+  for (const functionName of Object.keys(zeroOutMap)) {
+    const myFunc = zeroOutMap[functionName];
+    it(`${functionName} does not leak memory`, () => {
+      // The first call to myFunc creates and keeps any internal singleton
+      // tensors.  Subsequent calls should always create exactly the expected
+      // count of new tensors.
+      myFunc();
+      const numTensorsBefore = memory().numTensors;
+      myFunc();
+      const numTensorsAfter = memory().numTensors;
+      expect(numTensorsAfter).toEqual(numTensorsBefore);
+    });
+  }
+});
+
+
+describe('Memory leak check.  Functions with 1 tensor output:', () => {
+  // Reusable arguments
+  const x1D = tensor1d([4, 3, 2, 1]);
+  const x2D = tensor2d([[4, 3], [2, 1]]);
+  const x3D = tensor3d([[[0]]]);
+  const x4D = tensor4d([[[[0]]]]);
+  const xScalarAs1D = tensor1d([1978]);
+  // Functions that should produce one new Tensor.
+  // tslint:disable-next-line:no-any
+  const oneOutMap: { [key: string]: () => any } = {};
+  oneOutMap['batchFlatten'] = () => K.batchFlatten(x2D);
+  oneOutMap['cast'] = () => K.cast(x1D, 'float32');
+  oneOutMap['concatenate'] = () => K.concatenate([x1D, x1D]);
+  oneOutMap['concatAlongFirstAxis'] =
+     () => K.concatAlongFirstAxis(x1D, x1D);
+  oneOutMap['expandDims'] = () => K.expandDims(x1D);
+  oneOutMap['flatten'] = () => K.flatten(x1D);
+  oneOutMap['repeat'] = () => K.repeat(x2D, 2);
+  oneOutMap['reshape'] = () => K.reshape(x1D, [1,4]);
+  oneOutMap['reverse'] = () => K.reverse(x1D, 0);
+  oneOutMap['sliceAlongAxis'] = () => K.sliceAlongAxis(x4D, 0, 1, 1);
+  oneOutMap['sliceAlongFirstAxis'] = () => K.sliceAlongFirstAxis(x4D, 0, 1);
+  oneOutMap['sliceAlongLastAxis'] = () => K.sliceAlongFirstAxis(x4D, 0, 1);
+  oneOutMap['spatial2dPadding'] =
+    () => K.spatial2dPadding(x4D);
+  oneOutMap['squeeze'] = () => K.squeeze(xScalarAs1D, 0);
+  oneOutMap['temporalPadding'] = () => K.temporalPadding(x3D);
+  oneOutMap['transpose'] = () => K.transpose(x2D);
+
+  for (const functionName of Object.keys(oneOutMap)) {
+    const myFunc = oneOutMap[functionName];
+    it(`${functionName} does not leak memory`, () => {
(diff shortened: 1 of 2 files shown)


src/backend/tfjs_backend_test.ts
```diff
@@ -27,6 +27,135 @@ import * as K from './tfjs_backend';
...
+  const zeroOutMap: { [key: string]: () => any } = {};
+  zeroOutMap['countParms'] = () => K.countParams(x1D);
+  zeroOutMap['ndim'] = () => K.ndim(x1D);
+  zeroOutMap['normalizeAxis'] = () => K.normalizeAxis(x1D, 0);
+  zeroOutMap['shape'] = () => K.shape(x1D);
+
+  for (const functionName of Object.keys(zeroOutMap)) {
+    const myFunc = zeroOutMap[functionName];
+    it(`${functionName} does not leak memory`, () => {
+      // The first call to myFunc may creat and keep any internal singleton
```
ericdnielsen: s/creat/create/

bileschi:
@Reviewers.  Please Hang on : I'm getting some new test failures.

bileschi:
Ok, this is ready for review now.

bileschi:
Review status: 0 of 2 files reviewed at latest revision, 1 unresolved discussion, some commit checks failed.

---

*[src/backend/tfjs_backend_test.ts, line 44 at r1](https://reviewable.io/reviews/tensorflow/tfjs-layers/166#-LC-KvtXAoTGNin1lrPd-r1-44:-LC-nGaq8AnUkanE2mmR:b-896fix) ([raw file](https://github.com/tensorflow/tfjs-layers/blob/dde49fe4d6d07d789c82aad6787bbb5a39751af2/src/backend/tfjs_backend_test.ts#L44)):*

s/creat/create/
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs-layers/166)*

caisq:
Review status: 0 of 2 files reviewed at latest revision, 1 unresolved discussion, some commit checks failed.

---

*[src/backend/tfjs_backend_test.ts, line 38 at r2](https://reviewable.io/reviews/tensorflow/tfjs-layers/166#-LC1YtU45eGZ0i6gQU6n:-LC1YtU45eGZ0i6gQU6o:biwfq4w) ([raw file](https://github.com/tensorflow/tfjs-layers/blob/4292baf18138055fe97448e8cced06a8fff2c5b3/src/backend/tfjs_backend_test.ts#L38)):*
> ```
> zeroOutMap['countParms'] = () => K.countParams(x1D);
> ```

Having a dictionary like this will introduce maintenance overhead. For example, when someone adds a new method, it is highly likely s/he will forget to do this, especially if it's an external contributor. 

Is there a better way to do this?

---


*Comments from [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs-layers/166)*

bileschi:
*[src/backend/tfjs_backend_test.ts, line 38 at r2](https://reviewable.io/reviews/tensorflow/tfjs-layers/166#-LC1YtU45eGZ0i6gQU6n:-LC3f1-a638qXjjSaDE5:b-axq0w2) ([raw file](https://github.com/tensorflow/tfjs-layers/blob/4292baf18138055fe97448e8cced06a8fff2c5b3/src/backend/tfjs_backend_test.ts#L38)):*

> ```
> zeroOutMap['countParms'] = () => K.countParams(x1D);
> ```

Having a dictionary like this will introduce maintenance overhead. For example, when someone adds a new method, it is highly likely s/he will forget to do this, especially if it's an external contributor. 

Is there a better way to do this?
</blockquote></details>

So here are the alternatives I've been able to come up with, in increasing order of preference.

A:  Do nothing, do not test for memory leaks
B:  Manual one-off testing for memory leaks, like what we see for QR
C:  Manually curating a list of functions to be tested, but centralized within one test (THIS)
D:  Automatically determining a list of functions to be tested.

I think you are asking for something like D?  There are two difficulties, though, as I see it:

D1:  Although one can determine the list of functions exported by the backend (K.keys()), it is difficult to determine the expected number of new tensors.  Can one programmatically determine this from the type information available at runtime?

D2:  One needs to call the function in order to measure the effect, but calling the function requires passing in valid arguments.  Automatically determining a set of valid arguments is difficult.

We could solve D1 by augmenting the type system to contain information about the number of additional tensors.

We could solve D2 by mandating that all exported functions include enough information to be able to call them without knowing their full signature.   Perhaps either with wrappers or additional attributes within the function.

Both of these solutions seem to be just as difficult as a registering list, as here.

Thoughts?  Are there better solutions for D1 and D2?

---


*Comments from [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs-layers/166)*

davidsoergel:
Review status: 0 of 2 files reviewed at latest revision, 2 unresolved discussions, some commit checks failed.

---

*[src/backend/tfjs_backend_test.ts, line 38 at r2](https://reviewable.io/reviews/tensorflow/tfjs-layers/166#-LC1YtU45eGZ0i6gQU6n:-LC4Xag76ed-y_41bVOu:bq9qdy2) ([raw file](https://github.com/tensorflow/tfjs-layers/blob/4292baf18138055fe97448e8cced06a8fff2c5b3/src/backend/tfjs_backend_test.ts#L38)):*

So here are the alternatives I've been able to come up with, in increasing order of preference.

A:  Do nothing, do not test for memory leaks
B:  Manual one-off testing for memory leaks, like what we see for QR
C:  Manually curating a list of functions to be tested, but centralized within one test (THIS)
D:  Automatically determining a list of functions to be tested.

I think you are asking for something like D?  There are two difficulties, though, as I see it:

D1:  Although one can determine the list of functions exported by the backend (K.keys()), it is difficult to determine the expected number of new tensors.  Can one programmatically determine this from the type information available at runtime?

D2:  One needs to call the function in order to measure the effect, but calling the function requires passing in valid arguments.  Automatically determining a set of valid arguments is difficult.

We could solve D1 by augmenting the type system to contain information about the number of additional tensors.

We could solve D2 by mandating that all exported functions include enough information to be able to call them without knowing their full signature.   Perhaps either with wrappers or additional attributes within the function.

Both of these solutions seem to be just as difficult as a registering list, as here.

Thoughts?  Are there better solutions for D1 and D2?


</blockquote></details>

How about augmenting describeMath{CPU, GPU, both} with an argument specifying how many tensors may be created?  Then memory leak testing can be built in to all of our other tests.

The argument should be optional so that existing tests work, but in this case it should log the actual number of tensors created in this case, to make it easy to update those tests.

WDYT?

---


*Comments from [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs-layers/166)*

bileschi:
*[src/backend/tfjs_backend_test.ts, line 38 at r2](https://reviewable.io/reviews/tensorflow/tfjs-layers/166#-LC1YtU45eGZ0i6gQU6n:-LC5JdVr2OrVni9sfHk0:b-a80w2g) ([raw file](https://github.com/tensorflow/tfjs-layers/blob/4292baf18138055fe97448e8cced06a8fff2c5b3/src/backend/tfjs_backend_test.ts#L38)):*

How about augmenting describeMath{CPU, GPU, both} with an argument specifying how many tensors may be created?  Then memory leak testing can be built in to all of our other tests.

The argument should be optional so that existing tests work, but in this case it should log the actual number of tensors created in this case, to make it easy to update those tests.

WDYT?
</blockquote></details>

I like the idea of incorporating the test into the framework, but I'm having trouble imagining how the syntax would work especially around separating those tensors created as arguments to the function, and the tensors created by the target function itself.

Note that we want to express the idea that, e.g.,  tf.add(x1, x2) creates exactly one new tensor, but only on the second invocation, and not counting the creation of x1 and x2

---


*Comments from [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs-layers/166)*

PR #167, opened 2018-05-08
Initial attempts at removing TensorInterface

ericdnielsen:
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs-layers/167)

### DIFF
--- a/src/backend/tfjs_backend_test.ts
+++ b/src/backend/tfjs_backend_test.ts
@@ -164,31 +164,31 @@ describeMathCPU('countParams', () => {
   it('Scalar', () => {
     const x = zeros([]);
     expect(K.countParams(x)).toEqual(1);
-    expect(K.countParams(new LayerVariable(x))).toEqual(1);
+    expect(K.countParams(new LayerVariable(x).read())).toEqual(1);
   });
 
   it('Tensor1D', () => {
     const x = zeros([3]);
     expect(K.countParams(x)).toEqual(3);
-    expect(K.countParams(new LayerVariable(x))).toEqual(3);
+    expect(K.countParams(new LayerVariable(x).read())).toEqual(3);
   });
 
   it('Tensor2D', () => {
     const x = zeros([3, 2]);
     expect(K.countParams(x)).toEqual(6);
-    expect(K.countParams(new LayerVariable(x))).toEqual(6);
+    expect(K.countParams(new LayerVariable(x).read())).toEqual(6);
   });
 
   it('Tensor3D', () => {
     const x = zeros([4, 3, 2]);
     expect(K.countParams(x)).toEqual(24);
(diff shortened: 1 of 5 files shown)


ericdnielsen:
This changes the pattern of "Tensor|TensorInterface" for things that can handle SymbolicTensors to "Tensor|SymbolicTensor".

A few places that I'm less happy about this conversion:
a) definitely check topology/wrapper changes -- we can't be 100% identical to the old code without the interface duck type.
b) the changes in backend test that need the explicit read() on all the LayerVariables.

src/engine/topology.ts
```diff
@@ -357,7 +357,7 @@ export abstract class Layer extends serialization.Serializable {
   protected _trainableWeights: LayerVariable[];
   private _nonTrainableWeights: LayerVariable[];
   private _losses: RegularizerFn[];
-  private _updates: TensorInterface[];
+  private _updates: Tensor[];
```
ericdnielsen: I think this is still correct, but it is more restrictive than what we had before.

src/backend/tfjs_backend_test.ts
```diff
@@ -164,31 +164,31 @@ describeMathCPU('countParams', () => {
   it('Scalar', () => {
     const x = zeros([]);
     expect(K.countParams(x)).toEqual(1);
-    expect(K.countParams(new LayerVariable(x))).toEqual(1);
+    expect(K.countParams(new LayerVariable(x).read())).toEqual(1);
```
ericdnielsen: LayerVariable didn't explicitly implement TensorInterface, but it did implicitly.

So option 1) as shown call .read() in these tests.
option 2) add LayerVariable to the type unions of Tensor|SymbolicTensor for the pure shape/dtype functions.

(If we take option 2, we probably want a export type TensorLike = Tensor|SymbolicTensor|LayerVariable; )

caisq:
:lgtm_strong:

---

Review status: 0 of 5 files reviewed at latest revision, all discussions resolved, some commit checks failed.

---

*[src/engine/topology.ts, line 360 at r1](https://reviewable.io/reviews/tensorflow/tfjs-layers/167#-LC-_XhJ9JTTMYPGV5Eu:-LC-_XhJ9JTTMYPGV5Ev:b-o4l5od) ([raw file](https://github.com/tensorflow/tfjs-layers/blob/101ba09af26c48381b0438ebce63005fc4962978/src/engine/topology.ts#L360)):*
> ```
> private _updates: Tensor[];
> ```

Can you add a comment here that _updates is current unused? It may get used in the future for things like dynamic (imperative) model definitions.

---


*Comments from [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs-layers/167#-:-LC-_jZ890YDG0zYHIV1:b-jbsf8z)*

PR #168, opened 2018-05-08
Initial sweep at pruning the -layers backend.

ericdnielsen:
We're aiming to removing as much as possible of the tfjs_backend file within the -layers repo.

Step 1: This PR removes all symbols in tfjs_backend that exactly forwarded to the core symbols.  Ie same function signature with no tweaking done to the inputs and outputs.

Step 2: Later PRs, will aim to converge layer's usage of functionally identical symbols -- some differ only in name (subtract versus sub), some differ only in unused arguments (dtype,shape are common extra args that -layers does nothing with and just calls the core symbol without them), some promote/demote from scalar<->single item 1d tensor.

Step 3: Review the remaining symbols, some might be candidates for migration to core, some migration to some other more appropriately named util/library in -layers.

One of the main points for review/discussion, convention on when/if any of these symbols should be imported directly, or only used off the imported top level symbol.  There's some mix and match on what felt most natural in code.  I think we'll want to come up with a more rigorous decision there though.

---
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs-layers/168)

### DIFF
--- a/src/activations.ts
+++ b/src/activations.ts
@@ -9,6 +9,7 @@
  */
 
 // Layer activation functions
+import * as tfc from '@tensorflow/tfjs-core';
 import {scalar, serialization, Tensor} from '@tensorflow/tfjs-core';
 
 import * as K from './backend/tfjs_backend';
@@ -74,13 +75,13 @@ export function elu(x: Tensor, alpha = 1): Tensor {
  * @returns Tensor with the same shape and dtype as `x`.
  */
 export function selu(x: Tensor): Tensor {
-  return K.selu(x);
+  return tfc.selu(x);
 }
 
 
 // Rectified linear unit
 export function relu(x: Tensor): Tensor {
-  return K.relu(x);
+  return tfc.relu(x);
 }
 
 /**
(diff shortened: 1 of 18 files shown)


caisq:
:lgtm_strong:

Thanks.

---

Review status: 0 of 18 files reviewed at latest revision, all discussions resolved.

---

*[src/backend/tfjs_backend_test.ts, line 2758 at r1](https://reviewable.io/reviews/tensorflow/tfjs-layers/168#-LC1_4qmAtCvThmwlrgn:-LC1_4qmAtCvThmwlrgo:b-b524m1) ([raw file](https://github.com/tensorflow/tfjs-layers/blob/38c15c9796794d6dc62f007d65aa1c8be1e01215/src/backend/tfjs_backend_test.ts#L2758)):*
> ```
> K.n
> ```

Maybe double-check you didn't remove any tests by mistake before merging this PR.

---


*Comments from [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs-layers/168#-:-LC1_ELn7uY03Q7rpvms:b-afugaw)*

PR #173, opened 2018-05-10
Add integration_test/tfjs2keras; Fix the bugs uncovered by the test

caisq:
This is a node.js-Python integration test. It checks that the models
exported from tfjs-layers can be loaded properly by Keras in Python.

Fix bugs uncovered by the test
- In Activation layers
- In RNN layers

Towards: https://github.com/tensorflow/tfjs/issues/13

---
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs-layers/173)


integration_tests/tfjs2keras/tfjs_save.ts
```diff
@@ -0,0 +1,142 @@
...
+      tfl.layers.dense({units: 100, inputShape: [200], activation: 'relu'}));
+  model.add(tfl.layers.dense({units: 50, activation: 'elu'}));
+  model.add(tfl.layers.dense({units: 24}));
+  model.add(tfl.layers.activation({activation: 'elu'}));
+  model.add(tfl.layers.dense({units: 8, activation: 'softmax'}));
+  fs.writeFileSync(exportPath, model.toJSON());
+}
+
+// Convolutional neural network (CNN).
+function exportCNNModel(exportPath: string): void {
```
ericdnielsen: Going to approve the PR as it is, but how hard would it be to evolve the system to use either directories in tfjs-examples or tfjs-models as the source of models rather than hard-coding them here.

PR #175, opened 2018-05-11
Use -core's DataType instead of -layers DType

ericdnielsen:
We had three definitions of 'data type' meaning things across core and layers.  Core has both a DType Enum and a DataType string-literal type union.  Layers had a DType Enum.  All co-existed be making sure the string keys of the enum and the type union matched.

There was definitely no reason at this point for Layers to have its own independent definition.  And after confirming with -core that DataType is the correct one, migrating all uses in Layers to the core symbol.

Core says they'll handle the required checking that a (JS rather than TS) string received is a valid value,  Layers shouldn't need to reproduce those checks.  I'm less sure how we want to handle functions that accept a subset of DataTypes (its common for 'bool' to be disallowed).  There's a slight mix of styles in this PR.  

A) dtype?: 'int32'|'float32' as the type (rather than DataType)
B) if (dtype == 'bool') throw

Generally speaking, I think I prefer option A, but I added more of B's since that matched the existing Layers pattern.  Let me know your thoughts.  Perhaps we need a DataTypeNumeric added to core to use in these places?  Or perhaps we're fine replicating the type literals.


bileschi:
Reviewed 27 of 27 files at r1.
Review status: all files reviewed at latest revision, all discussions resolved.

---

*[src/initializers_test.ts, line 28 at r1](https://reviewable.io/reviews/tensorflow/tfjs-layers/175#-LCFLKknETblCGSkWXBk:-LCFLKknETblCGSkWXBl:b-1zm0xs) ([raw file](https://github.com/tensorflow/tfjs-layers/blob/bb0d1b02ffd2381f86b88043e12c347b1004c2b2/src/initializers_test.ts#L28)):*
> ```
> 'float32'
> ```
Are these not represented as constants somewhere in tfjs-core?  If not, would that be desirable?  It seems like mis-typing one of these is a risk?

---


*Comments from [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs-layers/175)*

ericdnielsen:
Review status: all files reviewed at latest revision, 1 unresolved discussion.

---

*[src/initializers_test.ts, line 28 at r1](https://reviewable.io/reviews/tensorflow/tfjs-layers/175#-LCFLKknETblCGSkWXBk:-LCFNvW-EZB2t1OnhVlJ:b-fwu6i6) ([raw file](https://github.com/tensorflow/tfjs-layers/blob/bb0d1b02ffd2381f86b88043e12c347b1004c2b2/src/initializers_test.ts#L28)):*

> ```
> 'float32'
> ```
Are these not represented as constants somewhere in tfjs-core?  If not, would that be desirable?  It seems like mis-typing one of these is a risk?


</blockquote></details>

This is the usual pattern when working with a string type union, at least as far as I've seen.

If you're working in TS, TS tooling will catch the typo.  If you're working in JS, core should have fucntions that validate the received string is in the type union.  Places where core doesn't have such would be considered a bug.

---


*Comments from [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs-layers/175)*

bileschi:
:lgtm_strong:

---

Review status: all files reviewed at latest revision, 1 unresolved discussion, some commit checks failed.

---



*Comments from [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs-layers/175#-:-LCTurzR2Erq-0xe3ipl:b-jbsf8z)*

caisq:
:lgtm_strong:

Thanks. Please resolve conflict, rebase on master, make sure all tests pass and merge the PR.

---

Review status: all files reviewed at latest revision, 1 unresolved discussion, some commit checks failed.

---



*Comments from [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs-layers/175#-:-LCTvy2REIQrz51ym13g:b-fpdl55)*

davidsoergel:
:lgtm_strong:

---

Reviewed 27 of 27 files at r1.
Review status: all files reviewed at latest revision, 1 unresolved discussion, some commit checks failed.

---



*Comments from [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs-layers/175#-:-LCTwm_H15q8AlBOiIxn:b-jbsf8z)*

PR #176, opened 2018-05-11
Change activations to classes to solve a minification issue

ericdnielsen:
Solving issue: https://github.com/tensorflow/tfjs/issues/288

Activations were raw functions and thus had no place to hang a hard-coded string to match the required symbol for cross language compatibility.

### DIFF
--- a/src/activations.ts
+++ b/src/activations.ts
@@ -13,44 +13,21 @@ import * as tfc from '@tensorflow/tfjs-core';
 import {scalar, serialization, Tensor} from '@tensorflow/tfjs-core';
 
 import * as K from './backend/tfjs_backend';
-import {ValueError} from './errors';
+import {deserializeKerasObject} from './utils/generic_utils';
+
 
 export type ActivationFn = (tensor: Tensor, axis?: number) => Tensor;
-/** @docinline */
-export type ActivationIdentifier = 'elu'|'hardsigmoid'|'linear'|'relu'|'relu6'|
-    'selu'|'sigmoid'|'softmax'|'softplus'|'softsign'|'tanh'|string;
 
-// TODO(cais): Consider switching arg type from string to Enum.
-export function getActivation(activationType: ActivationIdentifier):
-    ActivationFn {
-  if (activationType == null) {
-    return linear;
-  } else if (activationType.toLowerCase() === 'elu') {
-    return elu;
-  } else if (activationType.toLowerCase() === 'hardsigmoid') {
-    return hardSigmoid;
-  } else if (activationType.toLowerCase() === 'linear') {
-    return linear;
-  } else if (activationType.toLowerCase() === 'relu') {
-    return relu;
-  } else if (activationType.toLowerCase() === 'relu6') {
-    return relu6;
-  } else if (activationType.toLowerCase() === 'selu') {
-    return selu;
-  } else if (activationType.toLowerCase() === 'sigmoid') {
-    return sigmoid;
-  } else if (activationType.toLowerCase() === 'softmax') {
-    return softmax;
-  } else if (activationType.toLowerCase() === 'softplus') {
-    return softplus;
-  } else if (activationType.toLowerCase() === 'softsign') {
-    return softsign;
-  } else if (activationType.toLowerCase() === 'tanh') {
-    return tanh;
-  } else {
-    throw new ValueError(`Unsupported activation function ${activationType}`);
+export abstract class Activation extends serialization.Serializable {
+  abstract apply(tensor: Tensor, axis?: number): Tensor;
+  getConfig(): serialization.ConfigDict {
+    return {};
   }
 }
+
+/** @docinline */
+export type ActivationIdentifier = 'elu'|'hardSigmoid'|'linear'|'relu'|'relu6'|
+    'selu'|'sigmoid'|'softmax'|'softplus'|'softsign'|'tanh'|string;
 
 /**
  * Exponential linear unit (ELU).
@@ -59,10 +36,13 @@ export function getActivation(activationType: ActivationIdentifier):
  * @param alpha: Scaling factor the negative section.
  * @return Output of the ELU activation.
  */
-export function elu(x: Tensor, alpha = 1): Tensor {
-  return K.elu(x, alpha);
+export class Elu extends Activation {
+  static readonly className = 'elu';
+  apply(x: Tensor, alpha = 1): Tensor {
+    return K.elu(x, alpha);
+  }
 }
-
+serialization.SerializationMap.register(Elu);
 
 /**
  * Scaled Exponential Linear Unit. (Klambauer et al., 2017).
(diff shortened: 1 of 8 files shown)


ericdnielsen:
I know I still need to clean up the JsDoc (switched from functions to classes, so formatting is off).

There's one oddity to this that warrants special attention -- the static className on the activations are defined as the python code function name (lowercase), as that's what's required for compatibility.  Elsewhere for all non-Activations, we use the upper case version.

davidsoergel:
Awesome, thanks for fixing this so quickly!  :lgtm_strong:

---

Review status: 0 of 8 files reviewed at latest revision, all discussions resolved.

---

*[src/activations.ts, line 186 at r2](https://reviewable.io/reviews/tensorflow/tfjs-layers/176#-LCFuToB7qEofD2NTsxF:-LCFuToB7qEofD2NTsxG:bb7phvt) ([raw file](https://github.com/tensorflow/tfjs-layers/blob/42ea11f23b3ae899a90db346c485a35759e543db/src/activations.ts#L186)):*
> ```XML
> }
> 
> export function deserializeActiviation(
> ```

typo

---

*[src/layers/core.ts, line 18 at r2](https://reviewable.io/reviews/tensorflow/tfjs-layers/176#-LCFv3Xf3SYWr9gHdXuN:-LCFv3Xf3SYWr9gHdXuO:b5oemls) ([raw file](https://github.com/tensorflow/tfjs-layers/blob/42ea11f23b3ae899a90db346c485a35759e543db/src/layers/core.ts#L18)):*
> ```
> ActivationFn
> ```
Can we ditch the rename to ActivationFn and just update the usages to Activation?

---


*Comments from [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs-layers/176#-:-LCFvGysD53WsZqJdTva:b-c70vfy)*

nsthorat:
You actually can hang a string off a vanilla function since it's an object: https://basarat.gitbooks.io/typescript/docs/types/callable.html

In TypeScript you'd write an interface:

```
interface Fn {
  (): Tensor;
  name: string;
}
```

In a JS console try this:

```
const fn = () => 1;
fn.test = 'hello';
console.log(fn()); // returns 1
console.log(fn.test); // returns 'hello'
```

---

Review status: 0 of 8 files reviewed at latest revision, 2 unresolved discussions, all commit checks successful.

---



*Comments from [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs-layers/176#-:-LCFw3M3AQWkFUY30oL7:b-8vmk6d)*

davidsoergel:
If you mean this:

    function elu(...)  {...}
    elu.className = 'elu';

I suspect that would integrate poorly with the rest of the serialization plumbing, which is already geared towards classes.  And anyway it seems error-prone: the name is mutable; someone implementing a new activation may overlook the className bit since it's not enforced via the Serializable interface; etc.

That said: I think we can we use this mechanism to avoid the apply() calls.  I.e., leave this PR as is, turning things into classes, but rename the apply() method to just '()' (producing the same effect as __apply__ in Python).  I looked for that option in response to Eric's question on chat about what to call the method, and just didn't find it.  Thanks for the tip!

---

Review status: 0 of 8 files reviewed at latest revision, 2 unresolved discussions, all commit checks successful.

---



*Comments from [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs-layers/176#-:-LCFzp_63KahHM3b1IIF:b-id4h1e)*

caisq:
Looking.

---

Review status: 0 of 8 files reviewed at latest revision, 2 unresolved discussions, some commit checks failed.

---



*Comments from [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs-layers/176#-:-LCTt99hB13HcY0iskeS:b3i089c)*

ericdnielsen:
I think I'd prefer avoiding the renaming apply() -> ().  That feels a little hacky to me, and if we've already established the convention in TFJS-Layers of using apply to on practically every other object we should either change all of them, or none of them to that pattern.

---

Review status: 0 of 8 files reviewed at latest revision, 2 unresolved discussions, some commit checks failed.

---

*[src/activations.ts, line 186 at r2](https://reviewable.io/reviews/tensorflow/tfjs-layers/176#-LCFuToB7qEofD2NTsxF:-LCTtQ9A2GxPxh0Z0OAb:b-896fix) ([raw file](https://github.com/tensorflow/tfjs-layers/blob/42ea11f23b3ae899a90db346c485a35759e543db/src/activations.ts#L186)):*

typo
</blockquote></details>

Done.

---

*[src/layers/core.ts, line 18 at r2](https://reviewable.io/reviews/tensorflow/tfjs-layers/176#-LCFv3Xf3SYWr9gHdXuN:-LCTtdZA9g0cEUmLm2mz:byy3xt8) ([raw file](https://github.com/tensorflow/tfjs-layers/blob/42ea11f23b3ae899a90db346c485a35759e543db/src/layers/core.ts#L18)):*

> ```
> ActivationFn
> ```
Can we ditch the rename to ActivationFn and just update the usages to Activation?
</blockquote></details>

There's an Activation Layer (line 399) so the symbol names collide which is why I renamed the import here. 

---


*Comments from [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs-layers/176#-:-LCTtzED9rM7kn8IpQMC:byutidm)*

caisq:
:lgtm_strong:

---

Review status: 0 of 8 files reviewed at latest revision, 2 unresolved discussions, some commit checks failed.

---

*[src/activations.ts, line 28 at r2](https://reviewable.io/reviews/tensorflow/tfjs-layers/176#-LCTuVlEEI5JSX7vLprp:-LCTuVlEEI5JSX7vLprq:b-ljo29b) ([raw file](https://github.com/tensorflow/tfjs-layers/blob/42ea11f23b3ae899a90db346c485a35759e543db/src/activations.ts#L28)):*
> ```
> {}
> ```

I might be missing something, but should this dict include the class name? 

---


*Comments from [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs-layers/176#-:-LCTv1uC8nIWnU9ZsAbH:b-jbsf8z)*

ericdnielsen:
Review status: 0 of 8 files reviewed at latest revision, 3 unresolved discussions, some commit checks failed.

---

*[src/activations.ts, line 28 at r2](https://reviewable.io/reviews/tensorflow/tfjs-layers/176#-LCTuVlEEI5JSX7vLprp:-LCTvHth4j4LEqEAUIp6:buzo10s) ([raw file](https://github.com/tensorflow/tfjs-layers/blob/42ea11f23b3ae899a90db346c485a35759e543db/src/activations.ts#L28)):*

> ```
> {}
> ```

I might be missing something, but should this dict include the class name? 
</blockquote></details>

The config of most objects doesn't include its class name.  In most cases there's an config wrapper:  outermostConfig = { className: Something, config: {}}.  This is returning the inner config ie the Config object that the constructor would accept, normally the constructor doesn't deal with its own className.

---


*Comments from [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs-layers/176)*

caisq:
Review status: 0 of 8 files reviewed at latest revision, 3 unresolved discussions, some commit checks failed.

---

*[src/activations.ts, line 28 at r2](https://reviewable.io/reviews/tensorflow/tfjs-layers/176#-LCTuVlEEI5JSX7vLprp:-LCTwA285sbw1ZS_ODcq:bb6sh2i) ([raw file](https://github.com/tensorflow/tfjs-layers/blob/42ea11f23b3ae899a90db346c485a35759e543db/src/activations.ts#L28)):*

The config of most objects doesn't include its class name.  In most cases there's an config wrapper:  outermostConfig = { className: Something, config: {}}.  This is returning the inner config ie the Config object that the constructor would accept, normally the constructor doesn't deal with its own className.
</blockquote></details>

Ack.

---


*Comments from [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs-layers/176)*

davidsoergel:
:lgtm_strong:

---

Reviewed 7 of 8 files at r1, 1 of 1 files at r3.
Review status: all files reviewed at latest revision, 3 unresolved discussions, some commit checks failed.

---



*Comments from [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs-layers/176#-:-LCTy2cH4er3G06fwnGJ:b-jbsf8z)*

PR #184, opened 2018-05-15
Move dedicated algorithms to their respective layers

ericdnielsen:
Continue moving more functions out of tfjs_backend.ts.  This moves the majority of high-level, special purpose math into the primary/solo layer that uses it.  (There's some minor sharing of some of the convolutional helpers between pooling, convolutional, and convolutional_depthwise, but I think the organization makes sense with pooling and depthwise importing from convolution for the helper).

This didn't touch the normalization helpers nor the loss/metric helpers as they seemed a little more entwined/more thought needed.

This also adopts the standard that most files should do:
import * as tfc from '@tensorflow/tfjs-core';
import {Scalar, scalar, serialization, Tensor, tensor2d, tensor3d, tensor4d} from '@tensorflow/tfjs-core';

Ie, we only directly import types, type constructors, and second-level imports from core.  Math/other functions we access off the tfc alias.  I think this strikes the right balance of avoiding "tfc." appearing every time we use Tensor/types from core, while still making it obvious when we're delegating calculation to core.

---
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs-layers/184)


caisq:
:lgtm_strong:

---

Review status: 0 of 13 files reviewed at latest revision, all discussions resolved, some commit checks failed.

---



*Comments from [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs-layers/184#-:-LCZLfPqDXxqGvhU9udH:b-jbsf8z)*

PR #185, opened 2018-05-15
Remove mean and argmax from layers backend

ericdnielsen:
Removes mean and argmax from layer's backend and changes code to directly use core's mean and argMax respectively. argMax had a different default axis in core versus layers, however all existing uses in layers explicitly passed a value rather than relying on the default.

### DIFF
--- a/src/backend/tfjs_backend.ts
+++ b/src/backend/tfjs_backend.ts
@@ -834,30 +834,6 @@ export function oneHot(indices: Tensor, numClasses: number): Tensor {
 /* Elementary math functions. */
 
 /**
- * Mean of a tensor, alongside the specified axis.
- * @param x: Input tensor.
- * @param axis: An integer or an Array of integers.
- * @param keepDims: A boolean, whether to keep the dimensions or not
- *   If `keepDims` is `false`, the rank of the tensor is reduced
- *   by `1 for each entry in `axis`. If `keepDims` is `true`,
- *   the reduced dimensions are retained with length 1.
- */
-export function mean(
-    x: Tensor, axis?: number|number[], keepDims?: boolean): Scalar|Tensor {
-  return tfc.mean(x, axis, keepDims);
-}
-
-/**
- * Returns the index of the maximum value along an axis.
- * @param x Input tensor.
- * @param axis Axis along which to perform the reduction.
- * @returns The argmax index tensor.
- */
-export function argmax(x: Tensor, axis = -1): Tensor {
-  return tfc.argMax(x, axis);
-}
-
-/**
  * Retrieves the elements of indices `indices` in the tensor `reference`.
  * @param reference A tensor.
  * @param indices An integer tensor of indices or an `Array` of integers.
--- a/src/backend/tfjs_backend_test.ts
+++ b/src/backend/tfjs_backend_test.ts
@@ -1013,59 +1013,6 @@ describeMathCPUAndGPU('OneHot', () => {
   });
 });
 
-describeMathCPUAndGPU('Mean', () => {
-  it('reduce_mean', () => {
-    expectTensorsClose(
-      K.mean(tensor2d([[4, -1], [0, -2]], [2, 2])), scalar(0.25));
-  });
-  it('mean 2D, axis=1, keepdims=false', () => {
-    expectTensorsClose(
-      K.mean(tensor2d([[4, -1], [0, -2]], [2, 2]), 1), tensor1d([1.5, -1]));
-  });
-  it('mean 2D, axis=1, keepdims=true', () => {
-    expectTensorsClose(
-      K.mean(tensor2d([[4, -1], [0, -2]], [2, 2]), 1, true),
-      tensor2d([[1.5], [-1]], [2, 1]));
-  });
-  it('mean 3D, axis=[1,2], keepdims=false', () => {
-    expectTensorsClose(
-      K.mean(
-        tensor3d([[[4, -1], [0, -2]], [[40, -10], [0, -20]]], [2, 2, 2]),
-        [1, 2]),
-      tensor1d([0.25, 2.5]));
-  });
-  it('mean 3D, axis=[1,2], keepdims=true', () => {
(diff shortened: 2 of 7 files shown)


nsthorat:
:lgtm_strong:

---

Review status: 0 of 7 files reviewed at latest revision, all discussions resolved.

---

*[src/losses.ts, line 36 at r1](https://reviewable.io/reviews/tensorflow/tfjs-layers/185#-LCZLICL4XpkjFJQXtct:-LCZLICL4XpkjFJQXtcu:b9w3bbw) ([raw file](https://github.com/tensorflow/tfjs-layers/blob/fa5443f0add6a7fde09219f2b590fb097077d53e/src/losses.ts#L36)):*
> ```XML
>  */
> export function meanSquaredError(yTrue: Tensor, yPred: Tensor): Tensor {
>   return tfc.mean(K.square(tfc.sub(yPred, yTrue)), -1);
> ```

while are you in these functions, could you tidy them? there are lots of tensors leaked from all of these. In general, the contract is that functions should produce no more tensors than it returns.

this would become:

```
return tfc.tidy(() => tfc.mean(K.square(tfc.sub(yPred, yTrue)), -1));
```

here and elsewhere where you can.

---


*Comments from [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs-layers/185#-:-LCZLZ4X8ofZUFXlxGME:b-jbsf8z)*

caisq:
:lgtm_strong:

---

Review status: 0 of 7 files reviewed at latest revision, 1 unresolved discussion, all commit checks successful.

---

*[src/losses.ts, line 36 at r1](https://reviewable.io/reviews/tensorflow/tfjs-layers/185#-LCZLICL4XpkjFJQXtct:-LCZMcz99o23IbLld52F:b-ksiunu) ([raw file](https://github.com/tensorflow/tfjs-layers/blob/fa5443f0add6a7fde09219f2b590fb097077d53e/src/losses.ts#L36)):*

while are you in these functions, could you tidy them? there are lots of tensors leaked from all of these. In general, the contract is that functions should produce no more tensors than it returns.

this would become:

```
return tfc.tidy(() => tfc.mean(K.square(tfc.sub(yPred, yTrue)), -1));
```

here and elsewhere where you can.
</blockquote></details>

I think we should be used the @operation decorator now? It does the same thing as tidy(), but lets you write more concise code.

---

*[src/backend/tfjs_backend.ts, line 857 at r1](https://reviewable.io/reviews/tensorflow/tfjs-layers/185#-LCZMFOv4AtTnhnBg4n2:-LCZMFOv4AtTnhnBg4n3:bgjsi59) ([raw file](https://github.com/tensorflow/tfjs-layers/blob/a304266533fb7e8faa81039c0f5818bfa277f00c/src/backend/tfjs_backend.ts#L857)):*
> ```
> axis
> ```

Core doc says this defaults to 0, whereas here we default to -1. 
https://js.tensorflow.org/api/0.10.0/#argMax

But since all internal uses by -layers specifies the axis explicitly. This should be fine.
(No action required.)

---


*Comments from [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs-layers/185#-:-LCZMlNkEsdYWzUKW-i2:b-jbsf8z)*

bileschi:
*[src/losses.ts, line 36 at r1](https://reviewable.io/reviews/tensorflow/tfjs-layers/185#-LCZLICL4XpkjFJQXtct:-LCZMTln7W5Gqb7dpP1f:b-bv87zx) ([raw file](https://github.com/tensorflow/tfjs-layers/blob/fa5443f0add6a7fde09219f2b590fb097077d53e/src/losses.ts#L36)):*

I think we should be used the @operation decorator now? It does the same thing as tidy(), but lets you write more concise code.
</blockquote></details>

We had a CL to do this, but the decision was to use @operation decorator instead.

We were waiting for the decorator to be exported from core in https://github.com/tensorflow/tfjs-core/commit/f746e4a7054155670c5ba8d420be58dae3a0fdc7

---


*Comments from [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs-layers/185)*

nsthorat:
Review status: 0 of 7 files reviewed at latest revision, 2 unresolved discussions, all commit checks successful.

---

*[src/losses.ts, line 36 at r1](https://reviewable.io/reviews/tensorflow/tfjs-layers/185#-LCZLICL4XpkjFJQXtct:-LCZNCbjEacX_Pc0ai4W:b-8nz2hw) ([raw file](https://github.com/tensorflow/tfjs-layers/blob/fa5443f0add6a7fde09219f2b590fb097077d53e/src/losses.ts#L36)):*

We had a CL to do this, but the decision was to use @operation decorator instead.

We were waiting for the decorator to be exported from core in https://github.com/tensorflow/tfjs-core/commit/f746e4a7054155670c5ba8d420be58dae3a0fdc7

</blockquote></details>

You can't use the operation decorator for floating functions, only methods, unfortunately. Is the idea to move these to static methods/

---


*Comments from [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs-layers/185)*

ericdnielsen:
Review status: 0 of 7 files reviewed at latest revision, 2 unresolved discussions, all commit checks successful.

---

*[src/losses.ts, line 36 at r1](https://reviewable.io/reviews/tensorflow/tfjs-layers/185#-LCZLICL4XpkjFJQXtct:-LCZQnpp1k6uPsITodQi:b-cct2iv) ([raw file](https://github.com/tensorflow/tfjs-layers/blob/fa5443f0add6a7fde09219f2b590fb097077d53e/src/losses.ts#L36)):*

You can't use the operation decorator for floating functions, only methods, unfortunately. Is the idea to move these to static methods/
</blockquote></details>

We've chatted, the plain is:
a) not tidy in this PR.
b) in a later (soon) PR, tidy/operation everything.  Prefer operation when possible, but don't change functions to methods just to use operation.

---


*Comments from [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs-layers/185)*

PR #195, opened 2018-05-17
Move losses helpers

ericdnielsen:
Moves the loss helpers out of backend to losses.  While inlining binaryCrossentropy, changes to always apply the mean.  Both usages (in losses and metrics) were always taking the mean, so it didn't seem worth finding a renaming of the internal non-mean symbol.

Also note, we had no usages of the non-default 'true' branch of the binaryCrossentropy, so that has also been removed.  If that's enough justification for re-introducing an internal only helper, let me know.

---
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs-layers/195)


caisq:
:lgtm_strong:

---

Review status: 0 of 5 files reviewed at latest revision, all discussions resolved, some commit checks failed.

---



*Comments from [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs-layers/195#-:-LCjbE9L-2uuMRWv4GxF:b-jbsf8z)*

PR #210, opened 2018-05-24
allow Request object for model loading

jameswex:
FEATURE

Implemented TODO that allows Request object in tf.loadModel, which allows for options (such as use of credentials) to be provided to the fetch call that loads the model files.

---
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs-layers/210)

### DIFF
--- a/src/models.ts
+++ b/src/models.ts
@@ -32,11 +32,13 @@ import {convertPythonicToTs} from './utils/serialization_utils';
  *  @param custom_objects Optional dictionary mapping names
  *       (strings) to custom classes or functions to be
  *       considered during deserialization.
+ *  @params requestOptions Optional init options for the model loading request.
  * @returns A TensorFlow.js Layers `Model` instance (uncompiled).
  */
 export async function modelFromJSON(
     modelAndWeightsConfig: ModelAndWeightsConfig,
-    customObjects?: serialization.ConfigDict): Promise<Model> {
+    customObjects?: serialization.ConfigDict,
+    requestOptions?: RequestInit): Promise<Model> {
   let modelTopology = modelAndWeightsConfig.modelTopology;
   if (modelTopology['model_config'] != null) {
     // If the model-topology JSON contains a 'model_config' field, then it is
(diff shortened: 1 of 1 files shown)


caisq:
Review status: 0 of 1 files reviewed at latest revision, all discussions resolved.

---

*[src/models.ts, line 41 at r1](https://reviewable.io/reviews/tensorflow/tfjs-layers/210#-LDOouK2BqGnSSYJpFLA:-LDOouK2BqGnSSYJpFLB:b-254dyk) ([raw file](https://github.com/tensorflow/tfjs-layers/blob/a486a10e54d1df499c17cdabca83437b6177d580/src/models.ts#L41)):*
> ```
>  requestOptions?: RequestInit)
> ```

James, thanks for sending this PR! I think adding requestOptions for HTTP request-based model loading makes a lot of sending, and we should definitely support it in TensorFlow.js. In fact, we already have such support for HTTP request-based *saving*. See the code at: 
https://github.com/tensorflow/tfjs-core/blob/master/src/io/browser_http.ts#L268

The class `BrowserHTTPRequest` in the above-mentioned file is where the request option should be added. After all, `tf.loadModel` supports loading not only via HTTP, but also via other routes such as from browser's IndexedDB, Local Storage. Soon we will add support for native file system in node.js as well. Request options does not apply in those non-HTTP routes. For more info on various support loading/saving routes, see: https://js.tensorflow.org/tutorials/model-save-load.html

The class `BrowserHTTPRequest' does not have the `load` method implemented yet. I plan to implement it soon. See the TODO item at: 
https://github.com/tensorflow/tfjs-core/blob/master/src/io/browser_http.ts#L107

If you are interested in working together with me on the task, you are more than welcome to collaborate, needless to say.

Thanks again!

---


*Comments from [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs-layers/210)*

davidsoergel:
Closing as obsolete.  (I'm pretty sure this was implemented in another PR around the same time).

PR #216, opened 2018-05-29
Config verification for Conv layers

bileschi:
Adds configuration checking for conv layers at construction time.  Performs some light refactoring to make `DepthwiseConv2D` descend from a new abstract ancestor `BaseConv` rather than `Conv2D`, as `Conv2D` includes required properties that `DepthwiseConv2D` doesn't use, such as `filters`.

FEATURE: Improved configuration error checking for conv layers

---
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs-layers/216)


caisq:
Review status: 0 of 6 files reviewed at latest revision, all discussions resolved, some commit checks failed.

---

*[src/layers/convolutional.ts, line 316 at r1](https://reviewable.io/reviews/tensorflow/tfjs-layers/216#-LDlCHVd4ANDL4Y7-Zez:-LDlCHVd4ANDL4Y7-Zf-:b-2qg7u5) ([raw file](https://github.com/tensorflow/tfjs-layers/blob/0e3aac9b6d032d3ea9e7b99301958e003e82651c/src/layers/convolutional.ts#L316)):*
> ```XML
>   protected readonly activation: Activation;
>   protected readonly useBias: boolean;
> ```

Add a comment here explaining why the bias-related members are here, but the kernel-related ones are not.

---

*[src/layers/convolutional.ts, line 351 at r1](https://reviewable.io/reviews/tensorflow/tfjs-layers/216#-LDlFyUqExd0_eV0IQel:-LDlFyUqExd0_eV0IQem:bw1ojdj) ([raw file](https://github.com/tensorflow/tfjs-layers/blob/0e3aac9b6d032d3ea9e7b99301958e003e82651c/src/layers/convolutional.ts#L351)):*
> ```
> static verifyConfig
> ```

Should this be private or protected?

---

*[src/layers/convolutional.ts, line 355 at r1](https://reviewable.io/reviews/tensorflow/tfjs-layers/216#-LDlGxlh21We7FTQZlTJ:-LDlGxlh21We7FTQZlTK:bhfr2h0) ([raw file](https://github.com/tensorflow/tfjs-layers/blob/0e3aac9b6d032d3ea9e7b99301958e003e82651c/src/layers/convolutional.ts#L355)):*
> ```
> (
> ```

Superfluous parentheses.

---

*[src/layers/convolutional.ts, line 371 at r1](https://reviewable.io/reviews/tensorflow/tfjs-layers/216#-LDlH87cEDFllwxCVRNM:-LDlH87cEDFllwxCVRNN:bls5m8z) ([raw file](https://github.com/tensorflow/tfjs-layers/blob/0e3aac9b6d032d3ea9e7b99301958e003e82651c/src/layers/convolutional.ts#L371)):*
> ```
>  protected readonly kernelInitializer?: Initializer;
> ```

Like above, add a comment here to explain why the kernel-related members are here, but the bias-related ones are not.

---

*[src/layers/convolutional.ts, line 509 at r1](https://reviewable.io/reviews/tensorflow/tfjs-layers/216#-LDlHRJn4fYQeBOatyFJ:-LDlHRJn4fYQeBOatyFK:b1cq3wt) ([raw file](https://github.com/tensorflow/tfjs-layers/blob/0e3aac9b6d032d3ea9e7b99301958e003e82651c/src/layers/convolutional.ts#L509)):*
> ```
> static verifyConfig
> ```

Same as above. Should this be private or protected? Same below.

---

*[src/layers/convolutional.ts, line 511 at r1](https://reviewable.io/reviews/tensorflow/tfjs-layers/216#-LDlHJf6BWFNybrAIO9p:-LDlHJf6BWFNybrAIO9q:b5o40t7) ([raw file](https://github.com/tensorflow/tfjs-layers/blob/0e3aac9b6d032d3ea9e7b99301958e003e82651c/src/layers/convolutional.ts#L511)):*
> ```
> (
> ```

Superfluous parentheses. Same below.

---

*[src/layers/convolutional_test.ts, line 309 at r1](https://reviewable.io/reviews/tensorflow/tfjs-layers/216#-LDlJ_zHAJtVc0URtsSr:-LDlJ_zHAJtVc0URtsSs:bhby6hm) ([raw file](https://github.com/tensorflow/tfjs-layers/blob/0e3aac9b6d032d3ea9e7b99301958e003e82651c/src/layers/convolutional_test.ts#L309)):*
> ```
> (/kernelSize/);
> ```

Maybe extend this regex a little to check the correct info about array length in the error message.

---

*[src/layers/convolutional_test.ts, line 596 at r1](https://reviewable.io/reviews/tensorflow/tfjs-layers/216#-LDlJsJV0wTDkACy8pW5:-LDlJsJV0wTDkACy8pW6:b1urm3s) ([raw file](https://github.com/tensorflow/tfjs-layers/blob/0e3aac9b6d032d3ea9e7b99301958e003e82651c/src/layers/convolutional_test.ts#L596)):*
> ```
> bad config 
> ```
Be more specific.

---

*[src/utils/generic_utils.ts, line 425 at r1](https://reviewable.io/reviews/tensorflow/tfjs-layers/216#-LDlHwY8BmY46lPXrUoL:-LDlHwY8BmY46lPXrUoM:b-jcfnr9) ([raw file](https://github.com/tensorflow/tfjs-layers/blob/0e3aac9b6d032d3ea9e7b99301958e003e82651c/src/utils/generic_utils.ts#L425)):*
> ```
> if
> ```

Say "if and only if" for completeness.

---

*[src/utils/generic_utils.ts, line 426 at r1](https://reviewable.io/reviews/tensorflow/tfjs-layers/216#-LDlI89s-7XBr8GxFjbj:-LDlI89s-7XBr8GxFjbk:bihnep0) ([raw file](https://github.com/tensorflow/tfjs-layers/blob/0e3aac9b6d032d3ea9e7b99301958e003e82651c/src/utils/generic_utils.ts#L426)):*
> ```
> least `minLength`.
> ```

with length `>= minLength` and '<= maxLength`.

---

*[src/utils/generic_utils.ts, line 430 at r1](https://reviewable.io/reviews/tensorflow/tfjs-layers/216#-LDlIjQI-0zmrP_IdFqP:-LDlIjQI-0zmrP_IdFqQ:b-jadw6z) ([raw file](https://github.com/tensorflow/tfjs-layers/blob/0e3aac9b6d032d3ea9e7b99301958e003e82651c/src/utils/generic_utils.ts#L430)):*
> ```
>  maybeArray
> ```

Since you expect this to be an array, calling it "maybeArray" is confusing. Maybe call it "input" or "x"?

---

*[src/utils/generic_utils.ts, line 432 at r1](https://reviewable.io/reviews/tensorflow/tfjs-layers/216#-LDlHm-C-YgBbKLm6Sjy:-LDlHm-D25H2eor8usMd:b5o40t7) ([raw file](https://github.com/tensorflow/tfjs-layers/blob/0e3aac9b6d032d3ea9e7b99301958e003e82651c/src/utils/generic_utils.ts#L432)):*
> ```
> (
> ```

Superfluous parentheses. Same below.

---

*[src/utils/generic_utils.ts, line 432 at r1](https://reviewable.io/reviews/tensorflow/tfjs-layers/216#-LDlJEZeC1SvoqPppG2E:-LDlJEZeC1SvoqPppG2F:b-rxrned) ([raw file](https://github.com/tensorflow/tfjs-layers/blob/0e3aac9b6d032d3ea9e7b99301958e003e82651c/src/utils/generic_utils.ts#L432)):*
> ```
> return (
> ```

Add a sanity check here for `minLength >= 0`.

---

*[src/utils/generic_utils.ts, line 435 at r1](https://reviewable.io/reviews/tensorflow/tfjs-layers/216#-LDlIttb5XEbVhxlAI3g:-LDlIttb5XEbVhxlAI3h:bhfr2h0) ([raw file](https://github.com/tensorflow/tfjs-layers/blob/0e3aac9b6d032d3ea9e7b99301958e003e82651c/src/utils/generic_utils.ts#L435)):*
> ```
> (
> ```

Superfluous parentheses.

---

*[src/utils/generic_utils_test.ts, line 258 at r1](https://reviewable.io/reviews/tensorflow/tfjs-layers/216#-LDlJ8XTAGLA_Yjc-PTZ:-LDlJ8XTAGLA_Yjc-PT_:b-fpzcw9) ([raw file](https://github.com/tensorflow/tfjs-layers/blob/0e3aac9b6d032d3ea9e7b99301958e003e82651c/src/utils/generic_utils_test.ts#L258)):*
> ```
>   it('checks lengths', () => {
>     // [1,2,3] is longer than 1.
> ```

Cover the edge case of `minLength === 0`.

---


*Comments from [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs-layers/216)*

bileschi:
Review status: 0 of 6 files reviewed at latest revision, 15 unresolved discussions, some commit checks failed.

---

*[src/layers/convolutional.ts, line 316 at r1](https://reviewable.io/reviews/tensorflow/tfjs-layers/216#-LDlCHVd4ANDL4Y7-Zez:-LDlQj7U35E8EVOG9nV3:b-896fix) ([raw file](https://github.com/tensorflow/tfjs-layers/blob/0e3aac9b6d032d3ea9e7b99301958e003e82651c/src/layers/convolutional.ts#L316)):*

Add a comment here explaining why the bias-related members are here, but the kernel-related ones are not.
</blockquote></details>

Done.

---

*[src/layers/convolutional.ts, line 351 at r1](https://reviewable.io/reviews/tensorflow/tfjs-layers/216#-LDlFyUqExd0_eV0IQel:-LDlS35I1rq4j6LCVSCM:b-896fix) ([raw file](https://github.com/tensorflow/tfjs-layers/blob/0e3aac9b6d032d3ea9e7b99301958e003e82651c/src/layers/convolutional.ts#L351)):*

> ```
> static verifyConfig
> ```

Should this be private or protected?
</blockquote></details>

Done.

---

*[src/layers/convolutional.ts, line 355 at r1](https://reviewable.io/reviews/tensorflow/tfjs-layers/216#-LDlGxlh21We7FTQZlTJ:-LDlS2ht3FdWX7HUjR04:b-896fix) ([raw file](https://github.com/tensorflow/tfjs-layers/blob/0e3aac9b6d032d3ea9e7b99301958e003e82651c/src/layers/convolutional.ts#L355)):*

> ```
> (
> ```

Superfluous parentheses.
</blockquote></details>

Done.

---

*[src/layers/convolutional.ts, line 371 at r1](https://reviewable.io/reviews/tensorflow/tfjs-layers/216#-LDlH87cEDFllwxCVRNM:-LDlUE337bAfoF5fu7Zx:b-896fix) ([raw file](https://github.com/tensorflow/tfjs-layers/blob/0e3aac9b6d032d3ea9e7b99301958e003e82651c/src/layers/convolutional.ts#L371)):*

> ```
>  protected readonly kernelInitializer?: Initializer;
> ```

Like above, add a comment here to explain why the kernel-related members are here, but the bias-related ones are not.
</blockquote></details>

Done.

---

*[src/layers/convolutional.ts, line 509 at r1](https://reviewable.io/reviews/tensorflow/tfjs-layers/216#-LDlHRJn4fYQeBOatyFJ:-LDlTrx-2lRDtznY10aE:b-896fix) ([raw file](https://github.com/tensorflow/tfjs-layers/blob/0e3aac9b6d032d3ea9e7b99301958e003e82651c/src/layers/convolutional.ts#L509)):*

> ```
> static verifyConfig
> ```

Same as above. Should this be private or protected? Same below.
</blockquote></details>

Done.

---

*[src/layers/convolutional.ts, line 511 at r1](https://reviewable.io/reviews/tensorflow/tfjs-layers/216#-LDlHJf6BWFNybrAIO9p:-LDlUDa7CJ3rKu4K6Yip:b-896fix) ([raw file](https://github.com/tensorflow/tfjs-layers/blob/0e3aac9b6d032d3ea9e7b99301958e003e82651c/src/layers/convolutional.ts#L511)):*

> ```
> (
> ```

Superfluous parentheses. Same below.
</blockquote></details>

Done.

---

*[src/layers/convolutional_test.ts, line 596 at r1](https://reviewable.io/reviews/tensorflow/tfjs-layers/216#-LDlJsJV0wTDkACy8pW5:-LDlVEXkBlik2uuDQpBN:b-896fix) ([raw file](https://github.com/tensorflow/tfjs-layers/blob/0e3aac9b6d032d3ea9e7b99301958e003e82651c/src/layers/convolutional_test.ts#L596)):*

> ```
> bad config 
> ```
Be more specific.
</blockquote></details>

Done.

---

*[src/utils/generic_utils.ts, line 425 at r1](https://reviewable.io/reviews/tensorflow/tfjs-layers/216#-LDlHwY8BmY46lPXrUoL:-LDlWGzzENRCNZ46GQwr:b-896fix) ([raw file](https://github.com/tensorflow/tfjs-layers/blob/0e3aac9b6d032d3ea9e7b99301958e003e82651c/src/utils/generic_utils.ts#L425)):*

> ```
> if
> ```

Say "if and only if" for completeness.
</blockquote></details>

Done.

---

*[src/utils/generic_utils.ts, line 426 at r1](https://reviewable.io/reviews/tensorflow/tfjs-layers/216#-LDlI89s-7XBr8GxFjbj:-LDlWLAx4Hd58Jax2QbX:b-896fix) ([raw file](https://github.com/tensorflow/tfjs-layers/blob/0e3aac9b6d032d3ea9e7b99301958e003e82651c/src/utils/generic_utils.ts#L426)):*

> ```
> least `minLength`.
> ```

with length `>= minLength` and '<= maxLength`.
</blockquote></details>

Done.

---

*[src/utils/generic_utils.ts, line 430 at r1](https://reviewable.io/reviews/tensorflow/tfjs-layers/216#-LDlIjQI-0zmrP_IdFqP:-LDlWZOLD4wDS25atKhO:b-896fix) ([raw file](https://github.com/tensorflow/tfjs-layers/blob/0e3aac9b6d032d3ea9e7b99301958e003e82651c/src/utils/generic_utils.ts#L430)):*

> ```
>  maybeArray
> ```

Since you expect this to be an array, calling it "maybeArray" is confusing. Maybe call it "input" or "x"?
</blockquote></details>

Done.

---

*[src/utils/generic_utils.ts, line 432 at r1](https://reviewable.io/reviews/tensorflow/tfjs-layers/216#-LDlHm-C-YgBbKLm6Sjy:-LDlWnRQ0HxitOJH1dzD:b-mpfrdr) ([raw file](https://github.com/tensorflow/tfjs-layers/blob/0e3aac9b6d032d3ea9e7b99301958e003e82651c/src/utils/generic_utils.ts#L432)):*

> ```
> (
> ```

Superfluous parentheses. Same below.
</blockquote></details>

Found one set.  more?

---

*[src/utils/generic_utils.ts, line 432 at r1](https://reviewable.io/reviews/tensorflow/tfjs-layers/216#-LDlJEZeC1SvoqPppG2E:-LDlX6ZxETOnv_LyiA1u:bmhozrg) ([raw file](https://github.com/tensorflow/tfjs-layers/blob/0e3aac9b6d032d3ea9e7b99301958e003e82651c/src/utils/generic_utils.ts#L432)):*
> 
What behavior is expected if minLength < 0?   Would you expect to throw?   I would expect it to pass vacuously, which it does now, no?

---


*Comments from [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs-layers/216)*

bileschi:
Review status: 0 of 6 files reviewed at latest revision, 15 unresolved discussions, some commit checks failed.

---

*[src/utils/generic_utils.ts, line 435 at r1](https://reviewable.io/reviews/tensorflow/tfjs-layers/216#-LDlIttb5XEbVhxlAI3g:-LDlb77k3K1kEg1jjUWR:b-896fix) ([raw file](https://github.com/tensorflow/tfjs-layers/blob/0e3aac9b6d032d3ea9e7b99301958e003e82651c/src/utils/generic_utils.ts#L435)):*

> ```
> (
> ```

Superfluous parentheses.
</blockquote></details>

Done.

---

*[src/utils/generic_utils_test.ts, line 258 at r1](https://reviewable.io/reviews/tensorflow/tfjs-layers/216#-LDlJ8XTAGLA_Yjc-PTZ:-LDlb6qy2dTipddAv4fY:b-896fix) ([raw file](https://github.com/tensorflow/tfjs-layers/blob/0e3aac9b6d032d3ea9e7b99301958e003e82651c/src/utils/generic_utils_test.ts#L258)):*

> ```
>   it('checks lengths', () => {
>     // [1,2,3] is longer than 1.
> ```

Cover the edge case of `minLength === 0`.
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs-layers/216)*

src/layers/convolutional.ts
```diff
@@ -365,18 +371,47 @@ export abstract class Conv extends Layer {
             `convolution, but received ${JSON.stringify(this.dilationRate)}`);
       }
     }
+  }
 
-    this.activation = getActivation(config.activation);
-    this.useBias = config.useBias == null ? true : config.useBias;
+  protected static verifyConfig(config: BaseConvLayerConfig) {
```
ericdnielsen: Why static, this seems a little odd to me.

src/layers/convolutional.ts
```diff
@@ -365,18 +371,47 @@ export abstract class Conv extends Layer {
...
+  // bias. Kernel-related properties are defined here rather than in the
+  // superclass because some convolution subclasses use different names and
+  // configuration parameters for their internal kernel state.
+  protected readonly kernelInitializer?: Initializer;
+  protected readonly kernelConstraint?: Constraint;
+  protected readonly kernelRegularizer?: Regularizer;
+
+  constructor(rank: number, config: ConvLayerConfig) {
+    super(rank, config as BaseConvLayerConfig);
+    Conv.verifyConfig(config);
```
ericdnielsen: this could be super.verifyConfig(config) I think, if non static. and would survive renamed/different hierarchy refactors more easily.

caisq:
:lgtm_strong:

Thanks!

---

Review status: 0 of 6 files reviewed at latest revision, 17 unresolved discussions, some commit checks failed.

---

*[src/layers/convolutional.ts, line 376 at r2](https://reviewable.io/reviews/tensorflow/tfjs-layers/216#-LDhbYmJ9N-fcfDuGBd_-r2-376:-LDlmORk1ALcQyXWmbHy:b-4cy86c) ([raw file](https://github.com/tensorflow/tfjs-layers/blob/9650e6fef8efa856c73a5229fac75a981034efa9/src/layers/convolutional.ts#L376)):*

Why static, this seems a little odd to me.
</blockquote></details>

+1 Makes more sense to make it non-static. Same below.

---

*[src/utils/generic_utils.ts, line 432 at r1](https://reviewable.io/reviews/tensorflow/tfjs-layers/216#-LDlJEZeC1SvoqPppG2E:-LDlmYtF9e-b27Znw7D6:bm465d5) ([raw file](https://github.com/tensorflow/tfjs-layers/blob/0e3aac9b6d032d3ea9e7b99301958e003e82651c/src/utils/generic_utils.ts#L432)):*

> 
What behavior is expected if minLength < 0?   Would you expect to throw?   I would expect it to pass vacuously, which it does now, no?
</blockquote></details>

It's a sanity check for the usage of this function itself, to make sure the caller is using it correctly.

Something like `assert(minLength >= 0); assert(maxLength >= minLength);`

---


*Comments from [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs-layers/216#-:-LDlmjJt4EsmIJddRgTX:b-afugb9)*

src/layers/convolutional.ts
bileschi: We could use a pattern like:

```ts
export class A extends Layer {
  constructor(config) {
    super(config);
    this.verifyConfig(config);
  }
  protected verifyConfig(config: Config) { return config.a > 0 }
}
export class B extends A {
  constructor(config) {
    super(config);
    this.verifyConfig(config);
  }
  protected verifyConfig(config: Config) {
    super.verifyConfig(config);
    return config.b > 0;
  }
}
```

However, this pattern would result in the `verifyConfig`s each being called twice.  Once because of the call path

```
B.constructor 
  super(config)
    this.verifyConfig(config)   // calls B.verifyConfig since `this` is a B
      super.verifyConfig(config)   // calls A.verifyConfig since `super` here is an A
```
and then again because of the call path
```
B.constructor 
  this.verifyConfig(config)  // calls B.verifyConfig since `this` is a B
      super.verifyConfig(config)   // calls A.verifyConfig since `super` here is an A
```

Is there a better pattern here?  Is it ok if the calls are made more than once?

src/layers/convolutional.ts
bileschi: Originally I had them non-static, but was bitten by the following behavior:

imagine class B inherits from class A, and both A and B have method named `verifyConfig`

A.constructor calls `this.verifyConfig(config)`
B.constructor calls `super(config)` and then `this.verifyConfig()`

when creating a new B, the value of `this` is of type B, even when referenced by A's constructor. 
 Thus, the `verifyConfig` defined in B is called twice and verifyConfig defined in `A` is not called at all.

Making these calls static replaces calls to `this.verifyConfig` with `A.verifyConfig`.   Here is the reference I used to get a handle on this:
https://javascript.info/class-inheritance

src/layers/convolutional.ts
bileschi: Another slightly different option is to call `super.verifyConfig` directly from the constructor.  I believe this still has problem of multiple calls.

```ts
export class A extends Layer {
  constructor(config) {
    super(config);
    super.verifyConfig(config);
    this.verifyConfig(config);
  }
  protected verifyConfig(config: Config) { return config.a > 0 }
}
export class B extends A {
  constructor(config) {
    super(config);
    super.verifyConfig(config)
    this.verifyConfig(config);
  }
  protected verifyConfig(config: Config) { return config.b > 0; }
}
```

bileschi:
Review status: 0 of 6 files reviewed at latest revision, 17 unresolved discussions, some commit checks failed.

---

*[src/utils/generic_utils.ts, line 432 at r1](https://reviewable.io/reviews/tensorflow/tfjs-layers/216#-LDlJEZeC1SvoqPppG2E:-LDlslwdCgA_c-A5ZMMO:b-896fix) ([raw file](https://github.com/tensorflow/tfjs-layers/blob/0e3aac9b6d032d3ea9e7b99301958e003e82651c/src/utils/generic_utils.ts#L432)):*

It's a sanity check for the usage of this function itself, to make sure the caller is using it correctly.

Something like `assert(minLength >= 0); assert(maxLength >= minLength);`

</blockquote></details>

Done.

---

*[src/utils/generic_utils.ts, line 432 at r1](https://reviewable.io/reviews/tensorflow/tfjs-layers/216#-LDlHm-C-YgBbKLm6Sjy:-LDlsmj78ehDodEoRZmw:b-896fix) ([raw file](https://github.com/tensorflow/tfjs-layers/blob/0e3aac9b6d032d3ea9e7b99301958e003e82651c/src/utils/generic_utils.ts#L432)):*

Found one set.  more?
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs-layers/216)*

src/layers/convolutional.ts
ericdnielsen: one other thought I had,  if keeping them static which it sounds like we have to do.  Would we prefer to add the veritfy calls to the factory methods and not the constructors?  (So we can fail early -- however it does decouple the invariant....

bileschi: Some of these abstract classes don't have factory methods.   Also, putting the call into the constructor gives us the hierarchical calls to the super.verifyConfig for free.  If we were to move the call into the factory we would have something like
```js
static leafClass(config): Layer {
    GrandParentClass.verifyConfig(config);
    ParentClass.verifyConfig(confg);
    LeafClass.verifyConfig(config);
    return new LeafClass(config);
}
```
no?

PR #230, opened 2018-06-08
Allow 2D pooling layers to take a number as the value of strides

caisq:
Fixes: https://github.com/tensorflow/tfjs/issues/344

BUG

---
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs-layers/230)

### DIFF
--- a/src/layers/pooling.ts
+++ b/src/layers/pooling.ts
@@ -256,7 +256,7 @@ export interface Pooling2DLayerConfig extends LayerConfig {
    *
    * If `null`, defaults to `poolSize`.
    */
-  strides?: [number, number];
+  strides?: number|[number, number];
 
   /** The padding type to use for the pooling layer. */
   padding?: PaddingMode;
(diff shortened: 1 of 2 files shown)


bileschi:
:lgtm_strong:

---

Review status: 0 of 2 files reviewed at latest revision, all discussions resolved.

---

*[src/layers/pooling.ts, line 259 at r1](https://reviewable.io/reviews/tensorflow/tfjs-layers/230#-LEW-f6P4Duncc9vJRPQ:-LEW-f6P4Duncc9vJRPR:b-1y1epz) ([raw file](https://github.com/tensorflow/tfjs-layers/blob/4a3303c9a8195047d7a30f66dbeb5db8fbb8512e/src/layers/pooling.ts#L259)):*
> ```XML
>    * If `null`, defaults to `poolSize`.
>    */
>   strides?: number|[number, number];
> ```

Just a question here, is an array containing a single value allowed?
should I be allowed to pass 2, [2], or [2, 2]?   Or are only 2, and [2, 2] allowed?

---


*Comments from [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs-layers/230#-:-LEW086a2S6CxEejQIA_:b-jbsf8z)*

caisq:
Review status: 0 of 2 files reviewed at latest revision, 1 unresolved discussion, all commit checks successful.

---

*[src/layers/pooling.ts, line 259 at r1](https://reviewable.io/reviews/tensorflow/tfjs-layers/230#-LEW-f6P4Duncc9vJRPQ:-LEZwckqAquqCCpP8Tem:b7qamqd) ([raw file](https://github.com/tensorflow/tfjs-layers/blob/4a3303c9a8195047d7a30f66dbeb5db8fbb8512e/src/layers/pooling.ts#L259)):*

Just a question here, is an array containing a single value allowed?
should I be allowed to pass 2, [2], or [2, 2]?   Or are only 2, and [2, 2] allowed?
</blockquote></details>

Good question. The answer is that an array of a single value is not allowed. The check on line 287 covers that. I added a unit test here to cover it as well.

---


*Comments from [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs-layers/230)*

PR #234, opened 2018-06-12
No-scheme relative path should be routed to HTTP handler only in browser

caisq:
* Currently, a call like tf.loadModel('./model.json') will be routed
  to tf.io.browserHTTPRequest regardless of whether the environment is
  browser or node.js. This will cause issues for node.js.
  This change makes it so that this routing of relative paths to
  browserHTTPRequest (using browser `fetch`) is done only in the
  browser. This implies that paths without a scheme is not supported.
  In other words, in node.js, a path given to tf.loadModel must have a
  scheme, e.g., tf.loadModel('file://./mode.json'),
  tf.loadModel('https://localhost/model.json').

Towards: https://github.com/tensorflow/tfjs/issues/343

---
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs-layers/234)

### DIFF
--- a/src/models.ts
+++ b/src/models.ts
@@ -203,7 +203,7 @@ export async function loadModelInternal(pathOrIOHandler: string|
     const handlers = io.getLoadHandlers(pathOrIOHandler);
     if (handlers.length === 0 && ENV.get('IS_BROWSER')) {
       // For backward compatibility: if no load handler can be found,
-      // assume it is a relative http path. This only applies to the
+      // assume it is a relative http path. This applies only to the
       // browser environment.
       handlers.push(io.browserHTTPRequest(pathOrIOHandler));
     }
--- a/src/models.ts
+++ b/src/models.ts
@@ -11,7 +11,7 @@
 /* Original source keras/models.py */
 
 // tslint:disable:max-line-length
-import {doc, io, Scalar, serialization, Tensor} from '@tensorflow/tfjs-core';
+import {doc, ENV, io, Scalar, serialization, Tensor} from '@tensorflow/tfjs-core';
 
 import * as K from './backend/tfjs_backend';
 import {History} from './callbacks';
@@ -201,14 +201,19 @@ export async function loadModelInternal(pathOrIOHandler: string|
                                         io.IOHandler): Promise<Model> {
   if (typeof pathOrIOHandler === 'string') {
     const handlers = io.getLoadHandlers(pathOrIOHandler);
-    if (handlers.length === 0) {
+    if (handlers.length === 0 && ENV.get('IS_BROWSER')) {
       // For backward compatibility: if no load handler can be found,
-      // assume it is a relative http path.
+      // assume it is a relative http path. This only applies to the
+      // browser environment.
       handlers.push(io.browserHTTPRequest(pathOrIOHandler));
-    } else if (handlers.length > 1) {
+    }
+    if (handlers.length > 1) {
       throw new ValueError(
           `Found more than one (${handlers.length}) load handlers for ` +
           `URL '${pathOrIOHandler}'`);
+    } else if (handlers.length === 0) {
+      throw new ValueError(
+          `Cannot find any load handlers for path string '${pathOrIOHandler}'`);
     }
     pathOrIOHandler = handlers[0];
   }
--- a/src/models.ts
+++ b/src/models.ts
@@ -212,8 +212,14 @@ export async function loadModelInternal(pathOrIOHandler: string|
           `Found more than one (${handlers.length}) load handlers for ` +
           `URL '${pathOrIOHandler}'`);
     } else if (handlers.length === 0) {
-      throw new ValueError(
-          `Cannot find any load handlers for path string '${pathOrIOHandler}'`);
+      let errorMsg =
+          `Cannot find any load handlers for path string '${pathOrIOHandler}'.`;
+      if (ENV.get('IS_NODE')) {
+        errorMsg +=
+            ' For Node.js, a URL scheme (e.g., file://, http://, https://)' +
+            'is required.';
+      }
+      throw new ValueError(errorMsg);
     }
     pathOrIOHandler = handlers[0];
   }
@@ -331,8 +337,8 @@ export class Sequential extends Model {
    * ```
    * @param layer Layer instance.
    *
-   * @exception ValueError In case the `layer` argument does not know its input
-   *   shape.
+   * @exception ValueError In case the `layer` argument does not know its
+   * input shape.
    * @exception ValueError In case the `layer` argument has multiple output
    *   tensors, or is already connected somewhere else (forbidden in
    *   `Sequential` models).


bileschi:
:lgtm_strong:

---

---



*Comments from [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs-layers/234#-:-LEpIOGZ8xxgj8Jha6YG:b-jbsf8z)*

dsmilkov:

---

*[src/models.ts, line 204 at r1](https://reviewable.io/reviews/tensorflow/tfjs-layers/234#-LEpvOT64_usIVpFAV27:-LEpvOT64_usIVpFAV28:bs80wvb) ([raw file](https://github.com/tensorflow/tfjs-layers/blob/7df94c5d608cd01307e25e349001124493e66472/src/models.ts#L204)):*
> ```XML
>   if (typeof pathOrIOHandler === 'string') {
>     const handlers = io.getLoadHandlers(pathOrIOHandler);
>     if (handlers.length === 0 && ENV.get('IS_BROWSER')) {
> ```

does it make sense to have a "default" handler? Default handler gets called for relative paths. So if in the browser, register the httpHandler as the default handler. If node, register the fileHandler as the default handler. cc @nsthorat for thoughts.

---

*[src/models.ts, line 218 at r1](https://reviewable.io/reviews/tensorflow/tfjs-layers/234#-LEpvJEh6pAgmTIxDm_k:-LEpvJEiDaWggBtNVxhc:b1hyc8k) ([raw file](https://github.com/tensorflow/tfjs-layers/blob/7df94c5d608cd01307e25e349001124493e66472/src/models.ts#L218)):*
> ```XML
>           `Cannot find any load handlers for path string '${pathOrIOHandler}'.`;
>       if (ENV.get('IS_NODE')) {
>         errorMsg +=
> ```

Can you add some unit tests that overwrite the IS_BROWSER ENV flag? 

---


*Comments from [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs-layers/234)*

nsthorat:

---

*[src/models.ts, line 204 at r1](https://reviewable.io/reviews/tensorflow/tfjs-layers/234#-LEpvOT64_usIVpFAV27:-LEpwWpQ3AU71lSCP_Mt:brhqij9) ([raw file](https://github.com/tensorflow/tfjs-layers/blob/7df94c5d608cd01307e25e349001124493e66472/src/models.ts#L204)):*
<details><summary><i>Previously, dsmilkov (Daniel Smilkov) wrote…</i></summary><blockquote>

does it make sense to have a "default" handler? Default handler gets called for relative paths. So if in the browser, register the httpHandler as the default handler. If node, register the fileHandler as the default handler. cc @nsthorat for thoughts.
</blockquote></details>

I think that might make sense, but would be good to weigh against the complexity of having two handlers registered for the same thing (introducing priority, etc).

Just a musing: I don't really see a 3rd place where a relative URL would be registered, so I don't have a strong negative reaction against the proposed change here.

---


*Comments from [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs-layers/234)*

nsthorat:

---

*[src/models.ts, line 204 at r1](https://reviewable.io/reviews/tensorflow/tfjs-layers/234#-LEpvOT64_usIVpFAV27:-LEq0iKxBbqN-3hgTNC3:b-9mtqy1) ([raw file](https://github.com/tensorflow/tfjs-layers/blob/7df94c5d608cd01307e25e349001124493e66472/src/models.ts#L204)):*

I think that might make sense, but would be good to weigh against the complexity of having two handlers registered for the same thing (introducing priority, etc).

Just a musing: I don't really see a 3rd place where a relative URL would be registered, so I don't have a strong negative reaction against the proposed change here.
</blockquote></details>

Sorry, I didn't see the description of the PR -- I don't think it's okay to force "file://" in node for relative paths.

---


*Comments from [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs-layers/234)*

caisq:
Thank you all for the reviews so far. Let me know your thoughts, @nkreeger 

Two options for node.js + tf.loadModel

1) Let relative paths without URL schemes be treated as paths on the file system
2) Always require URL schemes in node.js + tf.loadModel. I.e., throw an error when there is no URL scheme.

This PR implements option 2.

---

---



*Comments from [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs-layers/234#-:-LEqDVJr6KlXa6Ioo71q:b-2gqixn)*

dsmilkov:
I'm definitely for #1 (Nikhil as well, based on his comment). Which is why I suggested an environment specific "default" handler that is responsible for handling relative paths.

---

---



*Comments from [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs-layers/234#-:-LEtTl9J1wY9CspDpHJb:b-p0lerm)*

nsthorat:
Yes -- I am also for #1.

---

---



*Comments from [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs-layers/234#-:-LEtV8oo0C6bR8A7DOnZ:b-mi25zp)*

bileschi:
This means that the same short program, executed in the two environments, will have different behavior.  This is not big problem, but we should be very clear about this re. documentation etc.

---

---



*Comments from [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs-layers/234#-:-LEtYsCy-qwxuKlk_lfF:brbnuov)*

nsthorat:
Yes, this is true, but this is how relative paths work in general in JavaScript, so it's not a divergence from expectations.

dsmilkov:
Hi Shanqing,

Just ran across this PR by chance. Let me know if you need more info/input from me to push this forward. Thanks!

caisq:
@dsmilkov Sorry - I backburnered this a little due to other things at hand. It's also because users of tfjs-node have at least one route of saving/loading working for them already, namely file://. But I do plan to get this done soon.

dsmilkov:
Makes sense - I trust your priority judgement :)

jafaircl:
Any update on this? tfjs@0.14.1 seems to have broken file:// loading

jafaircl:
@bileschi I looked on the js.tensorflow.org site [here](https://js.tensorflow.org/api/0.14.1/#Models-Loading)

bileschi:
@tafsiri Does it make sense to add node-only API documentation here?  It seems to me the answer is yes, but it would be nice to have some sort of tag to indicate when API calls won't be available in some environments.

@nkreeger thoughts?

See also @jafaircl 's response on stackoverflow here : https://stackoverflow.com/questions/53031887/unable-to-load-python-trained-model-in-tensorflow-js

tafsiri:
@bileschi From what I can gather this would just be documenting the behavior of the function (`tf.loadModel`) in node vs the browser? I think that makes sense here.

In cases where the function would not be available in a particular environment i believe the intent was to put it in an environment namespace like `tf.node` or `tf.browser` but i'm not sure if that applies here.

caisq:
Since this is not super high priority. I'll close the PR and re-open it later in the monorepo.

PR #239, opened 2018-06-15
Add dist-es6/ to .gitignore

caisq:
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs-layers/239)


caisq:
Discussed with @nsthorat offline. There should no longer be any dist-es6/

PR #245, opened 2018-06-20
[WIP] Fix bug in `dilationRate` of `Conv1d` in a 'model.json'.

jusjusjus:
#### Description
This pull request normalizes treatment of `strides` and `dilationRate` for `Conv1d` layers.

* Expanded unit test 'conv1d' to cover `dilationRate=2` with `stride=1`.
---
##### For repository owners only:

Please remember to apply all applicable tags to your pull request.
Tags: FEATURE, BREAKING, BUG, PERF, DEV, DOC, SECURITY

For more info see: https://github.com/tensorflow/tfjs/blob/master/DEVELOPMENT.md

PR #248, opened 2018-06-21
Make sure dilationRate is a number for 1D convolutions

dfoody:
BUG

#### Description
When a model is exported from Keras, 1D convolutions contain a dilation rate as an array of length 1. Code in a number of places, however, assumes that the dilationRate is either a number (for 1d convolutions) or an array of length 2. While some places support an array of length 1, others do not. For example, the dilationWidth (which is normally a number) becomes an array of length 1. This causes tfjs-node to fail as it passes the dilationWidth through to native code (as the native code expects dilationWidth to be a number).

The fix here is to convert a dilationRate of length 1 to a number as the layer is instantiated.

---

### DIFF
--- a/src/layers/convolutional.ts
+++ b/src/layers/convolutional.ts
@@ -354,13 +354,20 @@ export abstract class BaseConv extends Layer {
     this.biasRegularizer = getRegularizer(config.biasRegularizer);
     this.activityRegularizer = getRegularizer(config.activityRegularizer);
     this.dilationRate = config.dilationRate == null ? 1 : config.dilationRate;
-    if (this.rank === 1 &&
-        (Array.isArray(this.dilationRate) &&
-         (this.dilationRate as number[]).length !== 1)) {
-      throw new ValueError(
-          `dilationRate must be a number or an array of a single number ` +
-          `for 1D convolution, but received ` +
-          `${JSON.stringify(this.dilationRate)}`);
+    if (this.rank === 1 && Array.isArray(this.dilationRate)) {
+      if ((this.dilationRate as number[]).length !== 1) {
+        throw new ValueError(
+            `dilationRate must be a number or an array of a single number ` +
+            `for 1D convolution, but received ` +
+            `${JSON.stringify(this.dilationRate)}`);
+      } else {
+        // The tensorflow native backend (via tfjs-node) does not support a
+        // dilationRate as an array of length 1 (which is what is output from
+        // a model converted from Keras). And, in fact, it appears the tfjs-core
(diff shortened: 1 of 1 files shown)


dfoody:
Hi,

I signed it!

- Dan Foody
 Sent via Cloze

dfoody:
@jusjusjus Yep, your PR looks like it would solve the issue as well (I hadn't seen it before I made the change I did). I have no particular preference of which PR gets accepted as long as the bug gets resolved!

caisq:
Review status: 0 of 1 LGTMs obtained

---

*[src/layers/convolutional.ts, line 350 at r1](https://reviewable.io/reviews/tensorflow/tfjs-layers/248#-LFbeTCO5lEnCD1uupqq:-LFbeTCO5lEnCD1uupqr:b-kc2zld) ([raw file](https://github.com/tensorflow/tfjs-layers/blob/094b293d78a40393b4676f6c26c5aebc3024fb42/src/layers/convolutional.ts#L350)):*
> ```
> == 1 && Array.isArray(this.dilationRate)) {
> ```

---

*[src/layers/convolutional.ts, line 369 at r1](https://reviewable.io/reviews/tensorflow/tfjs-layers/248#-LFbgCoeCXwmDhmM4dGy:-LFbgCoeCXwmDhmM4dGz:b-gags5k) ([raw file](https://github.com/tensorflow/tfjs-layers/blob/e97836a55eab695b26bfaeb112219db3860fda06/src/layers/convolutional.ts#L369)):*
> ```
> this.dilationRate
> ```

First of all, thank you for sending the PR! Below is my comment.

Please see the code on line 509. Given this change, `dilationRate` will be serialized as a single number for `Conv1D` layers. That would be inconsistent with Python Keras, which always serializes `dilation_rate` of `Conv1D` as a length-1 array.

I wonder whether it suffices to just change the call to the tfjs-core function on line 457. There, when it calls `conv1DwithBias`, if `this.dilationRate` is an array, just take the first element.

---


*Comments from [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs-layers/248)*

dfoody:
@caisq If a number (instead of a length-1 array) is provided for `dilationRate` should the `getConfig` return the original value provided, or standardize `dilationRate` to a length-1 array?

If it should standardize it to a length-1 array then the PR by @jusjusjus (https://github.com/tensorflow/tfjs-layers/pull/245) would be the best route to go I suspect (instead of my PR).

Alternately, if you'd prefer to keep dilationRate in the format it was originally provided, then I can go ahead and update my PR with your suggestion.

Thoughts?

PR #250, opened 2018-06-22
Refactor to remove circular imports.

davidsoergel:
BUG

Refactor to remove circular imports.  These imports have been working so far, but we prefer to enforce non-circularity via a lint rule.  Making this possible requires a number of refactorings, combined here into a single PR.

* Move SymbolicTensor from types.ts to topology.ts
* Factor out unique ID generation etc. to backend/state.ts
* Remove some trivial functions from tfjs_backend.ts and replace their usages with direct calls to tfjs-core (or to the existing backend/common.ts).
* Factor engine/logs.ts out from callbacks.ts
* Factor types_utils.ts out from generic_utils.ts
* Move gradients function from tfjs_backend.ts to variables.ts

---
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs-layers/250)


caisq:
Review status: 0 of 1 LGTMs obtained

---

*[src/engine/topology.ts, line 101 at r1](https://reviewable.io/reviews/tensorflow/tfjs-layers/250#-LFr8Q4BAEZzniFPHe0U:-LFr8Q4C0epZr6YlTvk8:b-o7fjak) ([raw file](https://github.com/tensorflow/tfjs-layers/blob/2be28786274c8e1363d04f8d072d1d42507a8e6e/src/engine/topology.ts#L101)):*
> ```
> export class SymbolicTensor {
> ```

topology.ts is already pretty big. Can we put this class in a separate files, like `engine/symbolic_tensor.ts`?

---

*[src/engine/training_test.ts, line 1757 at r1](https://reviewable.io/reviews/tensorflow/tfjs-layers/250#-LFr9ZJ6CxZ7AXjUv3JZ:-LFr9ZJ6CxZ7AXjUv3J_:bacqfyp) ([raw file](https://github.com/tensorflow/tfjs-layers/blob/2be28786274c8e1363d04f8d072d1d42507a8e6e/src/engine/training_test.ts#L1757)):*
> ```
> describe('CallbackList', () => {
> ```

The same test isn't removed from its original location, is it? Why copy the test here?

---

*[src/utils/types_utils.ts, line 93 at r1](https://reviewable.io/reviews/tensorflow/tfjs-layers/250#-LFr8pAJAw4o5gE6dy0O:-LFr8pAJAw4o5gE6dy0P:b-bjrmwu) ([raw file](https://github.com/tensorflow/tfjs-layers/blob/2be28786274c8e1363d04f8d072d1d42507a8e6e/src/utils/types_utils.ts#L93)):*
> ```
> countParamsInWeights(weights: LayerVariable[]):
> ```

This method doesn't seem to have anything to do with "types".  Can we leave it where it was?

---


*Comments from [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs-layers/250)*

src/backend/tfjs_backend.ts
```diff
@@ -409,51 +344,6 @@ export function tile(x: Tensor, n: number|number[]): Tensor {
...
-  return new LayerVariable(tfc.eye(size, size, null, dtype), dtype, name);
-}
-
-/**
- * Multiply a scalar with an Tensor.
- * @param c The Scalar.
- * @param x The Tensor.
- * @returns The result of the multiplication.
- */
-export function scalarTimesArray(c: Scalar, x: Tensor): Tensor {
```
ericdnielsen: In discussion before on the backend cleanup, I thought out consensus was we wanted to keep the stricter type-checks of (Scalar,Tensor)->Tensor and not replace with the less type safe raw backend call.  Same decision applied to square() as well as both scalar(Times|Plus)Array

src/callbacks.ts
```diff
@@ -61,7 +43,7 @@ export abstract class Callback {
   // TODO(michaelterry): This type is a best guess.
   validationData: Tensor|Tensor[] = null;
   /** Instance of `keras.models.Model`. Reference of the model being trained. */
-  model: Model = null;
+  model: Container = null;
```
ericdnielsen: All the Model->Container changes here have me a tiny bit worried as we're changing the API contract.

src/constraints.ts
```diff
@@ -14,15 +14,17 @@
 import * as tfc from '@tensorflow/tfjs-core';
 import {doc, serialization, Tensor, tidy} from '@tensorflow/tfjs-core';
 
-import * as K from './backend/tfjs_backend';
+import {epsilon} from './backend/common';
+import {getScalar} from './backend/state';
+// import * as K from './backend/tfjs_backend';
```
ericdnielsen: Remove commented code.

src/engine/topology.ts
```diff
@@ -13,17 +13,20 @@
 // tslint:disable:max-line-length
 import {DataType, doc, Scalar, serialization, Tensor, tidy, util} from '@tensorflow/tfjs-core';
 
-import * as K from '../backend/tfjs_backend';
+// import * as K from '../backend/tfjs_backend';
```
ericdnielsen: Remove commented code.

src/types.ts
```diff
@@ -11,81 +11,17 @@
 /** Defines allowable data types for tensors. */
 
 // tslint:disable:max-line-length
-import {DataType, doc, Scalar, Tensor} from '@tensorflow/tfjs-core';
+import {Scalar, Tensor} from '@tensorflow/tfjs-core';
 
-import {getScopedTensorName, getUniqueTensorName} from './common';
-import {Layer} from './engine/topology';
+// import {Layer} from './engine/topology';
```
ericdnielsen: Remove commented code.

davidsoergel:
OK, I can still break the cycles by moving scalarTimesArray(), square(), etc. from tfjs_backend.ts to a new file (perhaps named core_math.ts or something).  @ericdnielsen @caisq @bileschi, lets discuss in person to be sure that's what we want before I revert all of the usages. Thanks!

---

Review status: 0 of 1 LGTMs obtained

---

*[src/callbacks.ts, line 46 at r1](https://reviewable.io/reviews/tensorflow/tfjs-layers/250#-LFd9Ljcw7LFU1olbtNm-r1-46:-LFrYThz6FqKzDPVo9o6:b-as34ak) ([raw file](https://github.com/tensorflow/tfjs-layers/blob/2be28786274c8e1363d04f8d072d1d42507a8e6e/src/callbacks.ts#L46)):*

All the Model->Container changes here have me a tiny bit worried as we're changing the API contract.
</blockquote></details>

It's true that this change is mostly a convenience for breaking a dependency cycle, so it feels a bit funny to broaden the API as a consequence.  I had thought it would be OK because nothing breaks in our code, but now I see that's because the model field is never used at all.  So, this change would indeed break anyone who writes a custom Callback that needs to trust that 'model' is really a Model.

Will revisit.

---

*[src/constraints.ts, line 19 at r1](https://reviewable.io/reviews/tensorflow/tfjs-layers/250#-LFd9Ljcw7LFU1olbtNo-r1-19:-LFr_uPm9Z_3PaxLkUoH:b-coqiy1) ([raw file](https://github.com/tensorflow/tfjs-layers/blob/2be28786274c8e1363d04f8d072d1d42507a8e6e/src/constraints.ts#L19)):*

Remove commented code.
</blockquote></details>

Oops thanks, done

---

*[src/types.ts, line 16 at r1](https://reviewable.io/reviews/tensorflow/tfjs-layers/250#-LFd9Ljejt3vfBiB7MtO-r1-16:-LFra0SZAd37A401fMnd:b-896fix) ([raw file](https://github.com/tensorflow/tfjs-layers/blob/2be28786274c8e1363d04f8d072d1d42507a8e6e/src/types.ts#L16)):*

Remove commented code.
</blockquote></details>

Done.

---

*[src/engine/topology.ts, line 16 at r1](https://reviewable.io/reviews/tensorflow/tfjs-layers/250#-LFd9Ljd_IhVoM8K9MkB-r1-16:-LFra6Jj05qtGZsfR5qw:b-896fix) ([raw file](https://github.com/tensorflow/tfjs-layers/blob/2be28786274c8e1363d04f8d072d1d42507a8e6e/src/engine/topology.ts#L16)):*

Remove commented code.
</blockquote></details>

Done.

---

*[src/engine/topology.ts, line 101 at r1](https://reviewable.io/reviews/tensorflow/tfjs-layers/250#-LFr8Q4BAEZzniFPHe0U:-LFra75Z9JE1sj6YGJ-2:b-8pvaqx) ([raw file](https://github.com/tensorflow/tfjs-layers/blob/2be28786274c8e1363d04f8d072d1d42507a8e6e/src/engine/topology.ts#L101)):*

> ```
> export class SymbolicTensor {
> ```

topology.ts is already pretty big. Can we put this class in a separate files, like `engine/symbolic_tensor.ts`?
</blockquote></details>

Unfortunately not, because Model and SymbolicTensor are mutually dependent.  Aside from the issue of having a large file (which is unfortunate, I agree) I think it actually makes sense to put these together because, semantically, SymbolicTensors are intimately tied to Layers.

---

*[src/engine/training_test.ts, line 1757 at r1](https://reviewable.io/reviews/tensorflow/tfjs-layers/250#-LFr9ZJ6CxZ7AXjUv3JZ:-LFrbl2K1gc3trKxnEP6:bxmi2w1) ([raw file](https://github.com/tensorflow/tfjs-layers/blob/2be28786274c8e1363d04f8d072d1d42507a8e6e/src/engine/training_test.ts#L1757)):*

> ```
> describe('CallbackList', () => {
> ```

The same test isn't removed from its original location, is it? Why copy the test here?
</blockquote></details>

Oops, thanks!  This was left over from an attempt to merge callbacks.ts with topology.ts, before I tried the Model->Container fix discussed above.  Obviously that made topology.ts even bigger :( but we'll see if I can detangle it another way.

---

*[src/utils/types_utils.ts, line 93 at r1](https://reviewable.io/reviews/tensorflow/tfjs-layers/250#-LFr8pAJAw4o5gE6dy0O:-LFrdhwC4RU6-b0Nltpg:bbi84h8) ([raw file](https://github.com/tensorflow/tfjs-layers/blob/2be28786274c8e1363d04f8d072d1d42507a8e6e/src/utils/types_utils.ts#L93)):*

> ```
> countParamsInWeights(weights: LayerVariable[]):
> ```

This method doesn't seem to have anything to do with "types".  Can we leave it where it was?
</blockquote></details>

We can't leave it there because it creates a circular dependency: src/utils/generic_utils.ts -> src/variables.ts -> src/constraints.ts -> src/utils/generic_utils.ts.  Putting it in layer_utils might have made sense, but creates another cycle.  So, just added variable_utils.

---


*Comments from [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs-layers/250#-:-LFsY4WV2Zahatpr8j9e:bsk56ds)*

davidsoergel:
Review status: 0 of 1 LGTMs obtained

---

*[src/callbacks.ts, line 46 at r1](https://reviewable.io/reviews/tensorflow/tfjs-layers/250#-LFd9Ljcw7LFU1olbtNm-r1-46:-LFsix5W22LLNU1ehNX0:blw0jr8) ([raw file](https://github.com/tensorflow/tfjs-layers/blob/2be28786274c8e1363d04f8d072d1d42507a8e6e/src/callbacks.ts#L46)):*

It's true that this change is mostly a convenience for breaking a dependency cycle, so it feels a bit funny to broaden the API as a consequence.  I had thought it would be OK because nothing breaks in our code, but now I see that's because the model field is never used at all.  So, this change would indeed break anyone who writes a custom Callback that needs to trust that 'model' is really a Model.

Will revisit.
</blockquote></details>

OK, I think we do need to retain the 'model: Model' field of Callback one way or another, so that leaves us two options. 

Option 1: (I did this here for discussion). Move the field to a new subclass, ModelAwareCallback, in training.ts.  This allows us to mostly leave callbacks.ts alone.  The main cost is that any user with a custom callback that needs the model field will have to extend ModelAwareCallback instead of Callback.  I.e., this is technically API-breaking but I think it's very minor.

Option 2: Just merge all of callbacks.ts into training.ts.  The file becomes large, but this correctly reflects that callbacks are intertwined with Model.

Preferences?

---


*Comments from [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs-layers/250)*

caisq:
Review status: 0 of 1 LGTMs obtained

---

*[src/callbacks.ts, line 46 at r1](https://reviewable.io/reviews/tensorflow/tfjs-layers/250#-LFd9Ljcw7LFU1olbtNm-r1-46:-LFsnA9A-p5e1g7nnu_3:b1s82c5) ([raw file](https://github.com/tensorflow/tfjs-layers/blob/2be28786274c8e1363d04f8d072d1d42507a8e6e/src/callbacks.ts#L46)):*

OK, I think we do need to retain the 'model: Model' field of Callback one way or another, so that leaves us two options. 

Option 1: (I did this here for discussion). Move the field to a new subclass, ModelAwareCallback, in training.ts.  This allows us to mostly leave callbacks.ts alone.  The main cost is that any user with a custom callback that needs the model field will have to extend ModelAwareCallback instead of Callback.  I.e., this is technically API-breaking but I think it's very minor.

Option 2: Just merge all of callbacks.ts into training.ts.  The file becomes large, but this correctly reflects that callbacks are intertwined with Model.

Preferences?
</blockquote></details>

I don't like either options. Maybe there is a third option: type model as `any` here. WDYT?

---


*Comments from [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs-layers/250)*

davidsoergel:
Review status: 0 of 1 LGTMs obtained

---

*[src/callbacks.ts, line 46 at r1](https://reviewable.io/reviews/tensorflow/tfjs-layers/250#-LFd9Ljcw7LFU1olbtNm-r1-46:-LFsnULG1FOPiQaCL4Uz:b-ab6b2) ([raw file](https://github.com/tensorflow/tfjs-layers/blob/2be28786274c8e1363d04f8d072d1d42507a8e6e/src/callbacks.ts#L46)):*

I don't like either options. Maybe there is a third option: type model as `any` here. WDYT?
</blockquote></details>

That's just an even more drastic version of what I did before, typing it as Container.  The issue is that then anyone writing a custom callback can't trust that it's really a Model.  If we think it' s OK to force them to cast (at least, the TS users) then we can just go back to the Container solution.  Sorry I neglected to include that in the list; let's call it Option 3 :)

---


*Comments from [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs-layers/250)*

ericdnielsen:
Review status: 0 of 1 LGTMs obtained

---

*[src/callbacks.ts, line 46 at r1](https://reviewable.io/reviews/tensorflow/tfjs-layers/250#-LFd9Ljcw7LFU1olbtNm-r1-46:-LFsnbEBAKY_641xC2Nj:b-1vqdmu) ([raw file](https://github.com/tensorflow/tfjs-layers/blob/2be28786274c8e1363d04f8d072d1d42507a8e6e/src/callbacks.ts#L46)):*

That's just an even more drastic version of what I did before, typing it as Container.  The issue is that then anyone writing a custom callback can't trust that it's really a Model.  If we think it' s OK to force them to cast (at least, the TS users) then we can just go back to the Container solution.  Sorry I neglected to include that in the list; let's call it Option 3 :)
</blockquote></details>

I vote very strongly against any.  We've gotten rid of most of the any's  I don't want to add new ones, especially ones that hack around a non-type issue.

---


*Comments from [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs-layers/250)*

caisq:
Review status: 0 of 1 LGTMs obtained

---

*[src/engine/topology.ts, line 101 at r1](https://reviewable.io/reviews/tensorflow/tfjs-layers/250#-LFr8Q4BAEZzniFPHe0U:-LFsnVu0AmJVoH3W564O:bzcpuc1) ([raw file](https://github.com/tensorflow/tfjs-layers/blob/2be28786274c8e1363d04f8d072d1d42507a8e6e/src/engine/topology.ts#L101)):*

Unfortunately not, because Model and SymbolicTensor are mutually dependent.  Aside from the issue of having a large file (which is unfortunate, I agree) I think it actually makes sense to put these together because, semantically, SymbolicTensors are intimately tied to Layers.
</blockquote></details>

Discussed offline. `SymbolicTensor` is used only in the context of the input and output of `Layers`. Therefore it makes logical sense to put `SymbolicTensor` here. So I'm fine with it as is.

---

*[src/engine/training_test.ts, line 1757 at r1](https://reviewable.io/reviews/tensorflow/tfjs-layers/250#-LFr9ZJ6CxZ7AXjUv3JZ:-LFsnlvj6PIm6CJFmXHD:b-wwxojo) ([raw file](https://github.com/tensorflow/tfjs-layers/blob/2be28786274c8e1363d04f8d072d1d42507a8e6e/src/engine/training_test.ts#L1757)):*

Oops, thanks!  This was left over from an attempt to merge callbacks.ts with topology.ts, before I tried the Model->Container fix discussed above.  Obviously that made topology.ts even bigger :( but we'll see if I can detangle it another way.
</blockquote></details>

Yeah, please detangle it before merging this PR. This is just a test, so it should be relatively straightforward to detangle it, right?

---

*[src/utils/types_utils.ts, line 93 at r1](https://reviewable.io/reviews/tensorflow/tfjs-layers/250#-LFr8pAJAw4o5gE6dy0O:-LFso9oH3ur63LFHQxZK:b-wd32lx) ([raw file](https://github.com/tensorflow/tfjs-layers/blob/2be28786274c8e1363d04f8d072d1d42507a8e6e/src/utils/types_utils.ts#L93)):*

We can't leave it there because it creates a circular dependency: src/utils/generic_utils.ts -> src/variables.ts -> src/constraints.ts -> src/utils/generic_utils.ts.  Putting it in layer_utils might have made sense, but creates another cycle.  So, just added variable_utils.
</blockquote></details>

Adding variable_utils.ts makes sense. But please remove the duplicate here.

---


*Comments from [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs-layers/250)*

davidsoergel:
Per offline discussion, we're now OK with removing the wrapper functions, so leaving this as is.

---

Review status: 0 of 1 LGTMs obtained

---

*[src/engine/training_test.ts, line 1757 at r1](https://reviewable.io/reviews/tensorflow/tfjs-layers/250#-LFr9ZJ6CxZ7AXjUv3JZ:-LFsryOjBghqQGfpiMRH:b8gfpr8) ([raw file](https://github.com/tensorflow/tfjs-layers/blob/2be28786274c8e1363d04f8d072d1d42507a8e6e/src/engine/training_test.ts#L1757)):*

Yeah, please detangle it before merging this PR. This is just a test, so it should be relatively straightforward to detangle it, right?
</blockquote></details>

This is entirely dependent on the Callback<->Model cycle discussion above.  Currently the History tests are here because History is a ModelAwareCallback, which in turn is necessary only because some tests expect it.  If we go the route of typing Callback.model as Container, then indeed these tests will move back to callbacks.ts.

---

*[src/utils/types_utils.ts, line 93 at r1](https://reviewable.io/reviews/tensorflow/tfjs-layers/250#-LFr8pAJAw4o5gE6dy0O:-LFssX4N6bubjRPVmIsW:b4srgxb) ([raw file](https://github.com/tensorflow/tfjs-layers/blob/2be28786274c8e1363d04f8d072d1d42507a8e6e/src/utils/types_utils.ts#L93)):*

Adding variable_utils.ts makes sense. But please remove the duplicate here.
</blockquote></details>

Sorry, what's duplicated?  I see only one copy of countParamsInWeights, in variable_utils.ts.

---


*Comments from [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs-layers/250#-:-LFssqTA83ckb-Slm4u2:b37por5)*

caisq:
Review status: 0 of 1 LGTMs obtained

---

*[src/callbacks.ts, line 46 at r1](https://reviewable.io/reviews/tensorflow/tfjs-layers/250#-LFd9Ljcw7LFU1olbtNm-r1-46:-LFtqZuC0gqfHx_r8pMN:bmmjjam) ([raw file](https://github.com/tensorflow/tfjs-layers/blob/2be28786274c8e1363d04f8d072d1d42507a8e6e/src/callbacks.ts#L46)):*

I vote very strongly against any.  We've gotten rid of most of the any's  I don't want to add new ones, especially ones that hack around a non-type issue.
</blockquote></details>

OK.  Let's ditch the `any` approach. I think adding a class "ModelAwareCallbacks" makes general sense. However, I'd like to suggest a slight improvement (hopefully) on that approach. The main problems with that approach are
1) It adds some code to `training.ts`. The file is already fairly large. I want to avoid adding more stuff (especially unrelated stuff) to it if possible.
2) It is API breaking as David pointed out.

How about this:
- Rename the `Callback` class to `BaseCallback` or `ModelUnwareCallback` in callbacks.ts. Rename callback.ts to base_callback.ts or something like that. `BaseCallback` can have `Container` type for its `model` member, thus breaking the circular dependency.
- The `Model` class in `training.ts` will refer to `BaseCallback`.
- Create a new file called `callbacks.ts`, and put essentially your `ModelAwareCallback` in that file. Of course, it should just be called `Callback` and be exported in the tf.* namespace.

I think this will address the two problems mentioned above. We avoid adding more stuff to `training.ts` and there will be no client-visible API change related to callbacks.

---

*[src/utils/types_utils.ts, line 93 at r1](https://reviewable.io/reviews/tensorflow/tfjs-layers/250#-LFr8pAJAw4o5gE6dy0O:-LFtYt4H2EZieWHNzqXQ:b-eh6e6q) ([raw file](https://github.com/tensorflow/tfjs-layers/blob/2be28786274c8e1363d04f8d072d1d42507a8e6e/src/utils/types_utils.ts#L93)):*
> countParamsInWeights

Ah - I misread the code. There is no duplicate now.

---


*Comments from [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs-layers/250)*

davidsoergel:
Review status: 0 of 1 LGTMs obtained

---

*[src/callbacks.ts, line 46 at r1](https://reviewable.io/reviews/tensorflow/tfjs-layers/250#-LFd9Ljcw7LFU1olbtNm-r1-46:-LFuApjwEBZZVgJSHVsQ:b-dybknn) ([raw file](https://github.com/tensorflow/tfjs-layers/blob/2be28786274c8e1363d04f8d072d1d42507a8e6e/src/callbacks.ts#L46)):*

OK.  Let's ditch the `any` approach. I think adding a class "ModelAwareCallbacks" makes general sense. However, I'd like to suggest a slight improvement (hopefully) on that approach. The main problems with that approach are
1) It adds some code to `training.ts`. The file is already fairly large. I want to avoid adding more stuff (especially unrelated stuff) to it if possible.
2) It is API breaking as David pointed out.

How about this:
- Rename the `Callback` class to `BaseCallback` or `ModelUnwareCallback` in callbacks.ts. Rename callback.ts to base_callback.ts or something like that. `BaseCallback` can have `Container` type for its `model` member, thus breaking the circular dependency.
- The `Model` class in `training.ts` will refer to `BaseCallback`.
- Create a new file called `callbacks.ts`, and put essentially your `ModelAwareCallback` in that file. Of course, it should just be called `Callback` and be exported in the tf.* namespace.

I think this will address the two problems mentioned above. We avoid adding more stuff to `training.ts` and there will be no client-visible API change related to callbacks.

</blockquote></details>

 This works great-- good thinking, thanks!  The only difference in my implementation is that BaseCallback has no model member at all.  It does have a setModel method that's a no-op.  

---


*Comments from [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs-layers/250)*

davidsoergel:
Review status: 0 of 1 LGTMs obtained

---

*[src/engine/training_test.ts, line 1757 at r1](https://reviewable.io/reviews/tensorflow/tfjs-layers/250#-LFr9ZJ6CxZ7AXjUv3JZ:-LFwXlYq95Z_PQlXTHaL:brj3gc4) ([raw file](https://github.com/tensorflow/tfjs-layers/blob/2be28786274c8e1363d04f8d072d1d42507a8e6e/src/engine/training_test.ts#L1757)):*

This is entirely dependent on the Callback<->Model cycle discussion above.  Currently the History tests are here because History is a ModelAwareCallback, which in turn is necessary only because some tests expect it.  If we go the route of typing Callback.model as Container, then indeed these tests will move back to callbacks.ts.
</blockquote></details>

Ah forgot to close this thread-- there are now no tests moved to training_test.ts.

---


*Comments from [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs-layers/250)*

caisq:
Review status: 0 of 1 LGTMs obtained

---

*[src/base_callbacks.ts, line 67 at r2](https://reviewable.io/reviews/tensorflow/tfjs-layers/250#-LFxMXbIABmpZJGshVgc:-LFxMXbIABmpZJGshVgd:br89cyk) ([raw file](https://github.com/tensorflow/tfjs-layers/blob/da9f51154ad9353a99916307bfbd3729e30555ce/src/base_callbacks.ts#L67)):*
> ```
> setModel(model: Container):
> ```

Can you add a comment here explaining the motivation for having this BaseCallback class and why `model` here is set to the type of `Container` instead of `Model`?

---

*[src/callbacks.ts, line 22 at r2](https://reviewable.io/reviews/tensorflow/tfjs-layers/250#-LFxN1ji2Arbmu6_hiaF:-LFxN1ji2Arbmu6_hiaG:b4i88b6) ([raw file](https://github.com/tensorflow/tfjs-layers/blob/da9f51154ad9353a99916307bfbd3729e30555ce/src/callbacks.ts#L22)):*
> ```
> Container
> ```

s/Container/Model/

---

*[src/variables.ts, line 311 at r2](https://reviewable.io/reviews/tensorflow/tfjs-layers/250#-LFxNQOVEd1XEFxy6rKm:-LFxNQOVEd1XEFxy6rKn:bkrhuzn) ([raw file](https://github.com/tensorflow/tfjs-layers/blob/da9f51154ad9353a99916307bfbd3729e30555ce/src/variables.ts#L311)):*
> ```
> * Control flow.
> ```

Why is this section called "control flow"?  `gradients` isn't exactly control flow.

---

*[src/engine/logs.ts, line 1 at r2](https://reviewable.io/reviews/tensorflow/tfjs-layers/250#-LFxNqoB60UUOkxRTTtj:-LFxNqoB60UUOkxRTTtk:b-csve5f) ([raw file](https://github.com/tensorflow/tfjs-layers/blob/da9f51154ad9353a99916307bfbd3729e30555ce/src/engine/logs.ts#L1)):*
> ```
> import {Scalar} from '@tenso
> ```

1) Add copyright header.
2) Why is this in `src/engine` instead of `src`? This file should be alongside and in the same directory as `base_callbacks.ts` and `callbacks.ts`, because the content is closely related.

---


*Comments from [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs-layers/250)*

davidsoergel:
Review status: 0 of 1 LGTMs obtained

---

*[src/base_callbacks.ts, line 67 at r2](https://reviewable.io/reviews/tensorflow/tfjs-layers/250#-LFxMXbIABmpZJGshVgc:-LFxSQap6D13NAgRGnBZ:b-896fix) ([raw file](https://github.com/tensorflow/tfjs-layers/blob/da9f51154ad9353a99916307bfbd3729e30555ce/src/base_callbacks.ts#L67)):*

> ```
> setModel(model: Container):
> ```

Can you add a comment here explaining the motivation for having this BaseCallback class and why `model` here is set to the type of `Container` instead of `Model`?
</blockquote></details>

Done.

---

*[src/callbacks.ts, line 22 at r2](https://reviewable.io/reviews/tensorflow/tfjs-layers/250#-LFxN1ji2Arbmu6_hiaF:-LFxSXoP8tifI_KXg-lV:bfv3b4d) ([raw file](https://github.com/tensorflow/tfjs-layers/blob/da9f51154ad9353a99916307bfbd3729e30555ce/src/callbacks.ts#L22)):*

> ```
> Container
> ```

s/Container/Model/
</blockquote></details>

Oops, good catch!

---

*[src/variables.ts, line 311 at r2](https://reviewable.io/reviews/tensorflow/tfjs-layers/250#-LFxNQOVEd1XEFxy6rKm:-LFxTEVNDpc3tAt1GwjR:b6ivfg5) ([raw file](https://github.com/tensorflow/tfjs-layers/blob/da9f51154ad9353a99916307bfbd3729e30555ce/src/variables.ts#L311)):*

> ```
> * Control flow.
> ```

Why is this section called "control flow"?  `gradients` isn't exactly control flow.
</blockquote></details>

Oh yeah I wondered about that and forgot to revisit.  This was just pasted in the move from tfjs_backend.js (https://github.com/tensorflow/tfjs-layers/blob/master/src/backend/tfjs_backend.ts#L912), but I just deleted it anyway.

---

*[src/engine/logs.ts, line 1 at r2](https://reviewable.io/reviews/tensorflow/tfjs-layers/250#-LFxNqoB60UUOkxRTTtj:-LFxTjLF6y6fU5rUSbzG:b-ml0l7r) ([raw file](https://github.com/tensorflow/tfjs-layers/blob/da9f51154ad9353a99916307bfbd3729e30555ce/src/engine/logs.ts#L1)):*

> ```
> import {Scalar} from '@tenso
> ```

1) Add copyright header.
2) Why is this in `src/engine` instead of `src`? This file should be alongside and in the same directory as `base_callbacks.ts` and `callbacks.ts`, because the content is closely related.
</blockquote></details>

1. Thanks for catching the missing header; done.
2. I figured it's also closely related to training.ts, but I'm happy either way (or rather, it's not clear to me that any of these subdirectories make sense, but that's another matter :)).  Just moved it to src.

---


*Comments from [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs-layers/250)*

caisq:
:lgtm_strong:

Thanks for doing this!

---

---



*Comments from [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs-layers/250#-:-LFxWUMf9PYPMe_8ZWlK:b-79wrs7)*

PR #255, opened 2018-06-29
Export LayerVariable from './variables'

nsthorat:
Fixes https://github.com/tensorflow/tfjs/issues/480

cc @asross

---
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs-layers/255)


davidsoergel:
:lgtm_strong:

---

---



*Comments from [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs-layers/255#-:-LGB1L7_Acs9jSWKfsnp:b-jbsf8z)*

ericdnielsen:
Was this intended to be a 'private'/implementation detail?

caisq:
I shared Eric's question/concern to a certain degree. But I'm not entirely against this PR.

@asross, I saw the issue you filed. https://github.com/tensorflow/tfjs/issues/480

Can you provide some more detail on why you need access to the `LayerVariable` type? I want to see if there is another way to do what you are trying to do. Thanks.

---

---



*Comments from [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs-layers/255#-:-LGB2Mbc-5lcsiwnJy5u:bv9wcan)*

nsthorat:
@asross is working on a federated learning experiment where we train a model client-side with either layers or a core model. This means his code operates on variables of a model (sending weight updates back to a server). He needs to be able to differentiate between a core variable and a layers variable.

caisq:
@nsthorat @asross Why do you need the `LayerVariable` type? Would it suffice to call the read() method of a `LayerVariable` to get the `Tensor`-type value?

caisq:
`Layer` objects also provide `getWeights()` and `setWeights()` methods, which are based on the `Tensor` type.

nsthorat:
He needs to know whether the variable is a core variable or a layers variable, before calling `read()` which does not exist on a core variable.

asross:
> Can you provide some more detail on why you need access to the LayerVariable type? I want to see if there is another way to do what you are trying to do. Thanks.

For federated learning, we need clients to be able to (1) serialize values of specific model variables and send them to the server, and (2) take updated weights from the server and assign them back to those variables. I think it might be possible to do this without interfacing with `LayerVariable` if we assume that we're always working with `tf.Model`s, but in Typescript, it forbids us from nicely isolating the logic to specific functions (which at the very least need to reference `LayerVariable` for argument typing).

Another reason we want to be able to access `LayerVariable` is that we want to support doing this kind of federated learning with lower-level models that rely directly on `tf.Variable`s. Since `LayerVariable` and `Variable` have different interfaces, that requires doing type-checks to decide whether to call, e.g., `write` or `assign`. That might be solvable changing the interface on `LayerVariable` so that you can call `assign` and `data`/`dataSync` (which would be aliases for `write` and `read().data`), but we'd still want `LayerVariable` for argument typing.

asross:
@caisq here are some examples of code that we currently have that demonstrates these requirements:
https://github.com/PAIR-code/federated-learning/blob/master/src/client/comm.ts#L149-L159
https://github.com/PAIR-code/federated-learning/blob/master/src/client/comm.ts#L60

asross:
> That might be solvable changing the interface on LayerVariable so that you can call assign and data/dataSync (which would be aliases for write and read().data), but we'd still want LayerVariable for argument typing.

Building on this argument, maybe we could have a `Variable` _interface_ that both `Variable` and `LayerVariable` support?

PR #261, opened 2018-07-07
Activate initialEpoch for model.fit

hpssjellis:
See feature request to activate initialEpoch at

[https://github.com/tensorflow/tfjs/issues/497](https://github.com/tensorflow/tfjs/issues/497)

---
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs-layers/261)

### DIFF
--- a/src/engine/training.ts
+++ b/src/engine/training.ts
@@ -1799,7 +1799,6 @@ export class Model extends Container implements tfc.InferenceModel {
     }
     return out;
     // TODO(cais): Add value to outLabels.
-    // TODO(cais): Add initialEpoch.
   }
 
   /**
--- a/src/engine/training.ts
+++ b/src/engine/training.ts
@@ -1308,7 +1308,7 @@ export class Model extends Container implements tfc.InferenceModel {
       batchSize?: number, epochs?: number, verbose?: number,
       callbacks?: BaseCallback[], valF?: (data: Tensor[]) => Scalar[],
       valIns?: Tensor[], shuffle?: boolean|string, callbackMetrics?: string[],
-      initialEpoch = 0, stepsPerEpoch?: number,
+      initialEpoch?: number, stepsPerEpoch?: number,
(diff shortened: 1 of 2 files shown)


hpssjellis:
@bileschi 

I coded this blind as I am not sure how to run the tests on my cloud version of tfjs-layers (It has problems launching chrome). If you look at the [code here](https://github.com/hpssjellis/tfjs-layers/blob/f05531800cc4c30a693b168af4345d532ea0d14e/src/engine/training_test.ts#L508-L519)

if initialEpoch is set to 1 then history.epoch should just be [1] instead of [0,1]

I hope.

hpssjellis:
@bileschi 

Quote from @shanqing-cai  

`initialEpoch?: number`
Why is it necessary to change = 0 to ?: number here? I don't think this change is necessary.

end of quote.

I can make the requested changes, I was just trying to make initialEpoch follow the same format as: epochs, batchSize and shuffle, since the following [code](https://github.com/hpssjellis/tfjs-layers/blob/master/src/engine/training.ts#L1308-L1324) already covers the default settings.

```
 if (batchSize == null) {
      batchSize = 32;
    }
    if (epochs == null) {
      epochs = 1;
    }
    if (shuffle == null) {
      shuffle = true;
    }
    if (initialEpoch == null) {
      initialEpoch = 0;
    }
```


P.S. I notice the above code does not detect incorrect negative inputs, would the following code be an improvement?

```
    if ((batchSize == null) || (batchSize < 0)) {
      batchSize = 32;
    }
    if ((epochs == null) || (epochs < 0)) {
      epochs = 1;
    }
    if (shuffle == null) {
      shuffle = true;
    }
    if ((initialEpoch == null) || (initialEpoch < 0)) {
      initialEpoch = 0;
    }
```

Either way I am fine with whatever you think is best, just thought I would explain my thinking with the changes.

PR #267, opened 2018-07-12
Remove the doc decorator. Split exports into multiple files.

nsthorat:
Given that decorators work quite poorly inside google, we've decided to remove it in favor of a typical jsdoc annotation. This gives us the added benefit of reducing the stack trace, and remove classes in exports.

Because we simply aliased namespaces like "layers" to the exports class, now we split each of the namespace exports into their own file and export the file as the namespace. This has the added benefit of making each of the export namespaces very clear.

Note that two methods were added to docs this change: `meanAbsoluteError` and `meanAbsolutePercentageError` which were not marked as static and thus were not usable (either by docs or by top-level function entry point).

This PR also moves some of the unit tests over to use the public API.

Before there were 83 methods, now 85:

```
Found: 
  6 headings
  21 subheadings
  85 methods
```

Corresponding parser change: https://github.com/tensorflow/tfjs-website/pull/180
Corresponding tfjs-core change: https://github.com/tensorflow/tfjs-core/pull/1158

This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs-layers/267)


nsthorat:
Thanks for the review!

PR #274, opened 2018-07-30
Add option yieldEvery to ModelFitConfig; add heuristics for auto yielding

caisq:
Previously, users have to call await tf.nextFrame() in the callbacks of Model.fit() calls in order
to ensure page responsiveness during long-running Model.fit() calls. 

This PR adds logic to automatically call await tf.nextFrame(). Specifically, a new field is added
to the config object for Model.fit(): `yieldEvery`.

- The default value of 'auto' lets TensorFlow.js measure the batch duration during the first
  several batches and calculate how many batches should pass before calling await
  tf.nextFrame() again.
- The value 'batch' and 'epoch' lets TensorFlow.js call await tf.nextFrame every batch or
  every epoch, respectively.
- The value 'never' disables await tf.nextFrame() calls. It is the legacy behavior. The user
  can still call await tf.nextFrame() in custom callbacks.

The 'auto' approach is a tradeoff between the need to ensure page responsiveness and
the need to prevent short-running Model.fit() calls from slowing down too much.

FEATURE

---
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs-layers/274)

### DIFF
--- a/src/base_callbacks.ts
+++ b/src/base_callbacks.ts
@@ -11,7 +11,7 @@
 /* Original source: keras/callbacks.py */
 
 // tslint:disable:max-line-length
-import {add, div, keep, mul, nextFrame, Scalar, Tensor, tidy} from '@tensorflow/tfjs-core';
+import {add, div, keep, mul, nextFrame, Scalar, Tensor, tidy, util} from '@tensorflow/tfjs-core';
 
 import {getScalar} from './backend/state';
 import {Container} from './engine/container';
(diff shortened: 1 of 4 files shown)


caisq:
Previously, users have to call await tf.nextFrame() in the callbacks of Model.fit() calls in order
to ensure page responsiveness during long-running Model.fit() calls.

This PR adds logic to automatically call await tf.nextFrame(). Specifically, a new field is added
to the config object for Model.fit(): yieldEvery.

The default value of 'auto' lets TensorFlow.js measure the batch duration during the first
several batches and calculate how many batches should pass before calling await
tf.nextFrame() again.
The value 'batch' and 'epoch' lets TensorFlow.js call await tf.nextFrame every batch or
every epoch, respectively.
The value 'never' disables await tf.nextFrame() calls. It is the legacy behavior. The user
can still call await tf.nextFrame() in custom callbacks.
The 'auto' approach is a tradeoff between the need to ensure page responsiveness and
the need to prevent short-running Model.fit() calls from slowing down too much.

FEATURE

PR #277, opened 2018-07-31
Enhance tfjs2keras integration test

caisq:
Enhance tfjs2keras integration test
    
Previously only the loading of model topology was tested.
    
This PR adds the test for the loading of weight values. In addition,
this PR lets each model run its predict() method using some
randomly generated data. The input data and the prediction() outputs
are saved to disk from TensorFlow.js. On the Python side, the
input tensors are loaded and used to invoke the loaded models.
The (near) equality of the outputs from Python and from those from
TensorFlow.js is asserted on for each model.
    
DEV

---
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs-layers/277)


caisq:
@googlebot Ignore CLA issue. It is me with a different email address.

PR #279, opened 2018-08-02
Factor out countTrainableParams()

davidsoergel:
Factor out and expose layer_utils/countTrainableParams(), because that's useful.

---
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs-layers/279)


nsthorat:
Correct, we should fix that everywhere. The one exception is `jasmine_util`, which depends on jasmine. If we exported this from `index.ts`, then we would bloat the standalone bundle with jasmine dependencies.

PR #281, opened 2018-08-03
Add Layer: Permute

caisq:
FEATURE

Fixes: https://github.com/tensorflow/tfjs/issues/577

---
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs-layers/281)

### DIFF
--- a/src/layers/core.ts
+++ b/src/layers/core.ts
@@ -12,12 +12,12 @@
  * TensorFlow.js Layers: Basic Layers.
  */
 
-import {Scalar, serialization, Tensor, tidy, util} from '@tensorflow/tfjs-core';
+import {Scalar, serialization, Tensor, transpose, tidy, util} from '@tensorflow/tfjs-core';
 
 import {Activation as ActivationFn, ActivationIdentifier, getActivation, serializeActivation} from '../activations';
 import * as K from '../backend/tfjs_backend';
 import {Constraint, ConstraintIdentifier, getConstraint, serializeConstraint} from '../constraints';
-import {Layer, LayerConfig} from '../engine/topology';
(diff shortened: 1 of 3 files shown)


src/layers/core.ts
```diff
@@ -607,3 +607,94 @@ export class Reshape extends Layer {
...
+          'Permute constructor requires `dims` to be an Array, but received ' +
+          `${config.dims} instead.`);
+    }
+
+    // Check the validity of the permutation indices.
+    const expectedSortedIndices = range(1, config.dims.length + 1);
+    if (!util.arraysEqual(config.dims.slice().sort(), expectedSortedIndices)) {
+      throw new Error(
+          'Invalid permutation `dims`: ' + JSON.stringify(config.dims) +
+          ' `dims` must contain consecutive integers starting from 1.');
```
bileschi: Nice error string :)
Maybe add the expected last index as well?  or how many were expected (i guess that's the same number anyway).

src/layers/core_test.ts
```diff
@@ -375,33 +373,33 @@ describeMathCPUAndGPU('Activation Layer: Tensor', () => {
 
   it('linear', () => {
     const x = mul(scalar(10), ones(inputShape));
-    const layer = new Activation({activation: 'linear'});
+    const layer = tfl.layers.activation({activation: 'linear'});
```
bileschi: Thanks for fixing these

src/layers/core.ts
bileschi: I was surprised that the batch dim is always skipped. may want to add something to that effect as well.  "(The batch dim is always unchanged by 'Permute')"

PR #282, opened 2018-08-06
Add new method dispose() to tf.layers.Layer and tf.Model

caisq:
FEATURE

This allows a layer or a model to release the memory (e.g., WebGL
textures) allocated for its weights.

Usage example:

```js
const model = tf.sequential();
model.add(tf.layers.dense({
  units: 10, activation: 'relu', inputShape: [20]}));
model.add(tf.layers.dense({units: 3, activation: 'softmax'}));
model.summary();

model.decRef();
// This disposes the four weights that belong to the model's layers.
```

We use reference counting because layers may be shared among multiple
models. Also, a model may be nested under other models.

When the reference counter of a layer or model decreases to zero, its
weights will be disposed. After that, the layer or model cannot be used
in `apply`, `predict`, `evaluate`, `fit` or `save` calls anymore.

Fixes: https://github.com/tensorflow/tfjs/issues/533

---
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs-layers/282)

### DIFF
--- a/src/engine/container.ts
+++ b/src/engine/container.ts
@@ -628,6 +628,35 @@ export abstract class Container extends Layer {
       outputShapes: this.outputs.map(x => x.shape)
     });
     this.built = true;
+    this._refCount = 1;  // The ref count of a container always start at 1.
+  }
+
+  protected assertNotDisposed() {
+    if (this._refCount === 0) {
+      throw new Error(`Container '${this.name}' is already disposed.`);
+    }
+  }
+
+  /**
+   * Decrease the reference count of the Container object by 1.
+   *
+   * A Container is reference-counted. Its reference count is incremented by 1
+   * when it is first constructed and when it is used as a Layer of another
+   * Container.
+   *
+   * If the reference count of a Container becomes 0, the `decRef` method of
+   * all its constituent Layer will be called.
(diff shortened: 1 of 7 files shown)


nsthorat:
dispose() actually does not guarantee a disposal of the underlying memory,
it just decrefs. If you reshape or clone a Tensor, we increase the refcount
to the underlying memory. You have to dispose both the original tensor and
the cloned tensor to actually free up the underlying memory. To me, this is
almost the same thing as what you are doing with model weights / layers.

nsthorat:
Sure, I’ll be back soon. We don’t mention it because it’s an implementation
detail and users should not have to worry about it.

In this case, it may be actually be important to the user and we should
document it.

caisq:
@nsthorat To answer your question, it is in container.ts under `class Container`. It is inherited by `Model` and `Sequential`.

caisq:
@dsmilkov Thanks for the comment, @dsmilkov . Let's discuss this in person when we meet.

nsthorat:
Just spoke to Daniel offline -- that approach unfortunately won't work because the layer itself is shared. Multiple models can both update the same weights.

PR #284, opened 2018-08-08
Bump version to 0.7.3

caisq:
Depend on tfjs-core 0.12.10

Also in this PR: Update a test related to 0-sizes 2D tensors, which used to be forbidden, but now is allowed, due to changes in tfjs-core.

---
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs-layers/284)


caisq:
Thanks for the review.

PR #285, opened 2018-08-09
Add test and guard for circular model weight assignment.

bileschi:
#### Description
Fixes tfjs#465 "Manually setting weights for mode" in the case of setting weights to weights the model already owns.  Previously the TFJS infrastructure was discarding needed weights.

Adds unit tests covering the behavior (fit model, assign weights to weights the model already owns, check the weights).

---

BUG: fixes tfjs-union#465

---
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs-layers/285)

### DIFF
--- a/src/engine/training_test.ts
+++ b/src/engine/training_test.ts
@@ -1086,6 +1086,36 @@ describeMathCPUAndGPU('Model.fit', () => {
         });
   });
 
+  it('and then set weights to new weights', async done => {
+    createDenseModelAndData(false, 'l1l2');
+    model.compile({optimizer: 'SGD', loss: 'meanSquaredError'});
+    model.fit(inputs, targets, {batchSize: numSamples, epochs: 2})
+        .then(history => {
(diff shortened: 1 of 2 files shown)


bileschi:
___
*[src/variables.ts, line 98 at r1](https://reviewable.io/reviews/tensorflow/tfjs-layers/285#-LJUvz797adc9kg9YAzK:-LJUxI1s2aKtH5HSf_rF:b-rdi903) ([raw file](https://github.com/tensorflow/tfjs-layers/blob/f0bbb0aa4159a984a9d5424efffc60499eb0af7a/src/variables.ts#L98)):*

> ```
>     if (this.val.id !== newVal.id) {
>       this.val.assign(newVal);
> ```

You can move the following if statement into this if statement, because if it is the same tensor, there is no need to apply the constraint again.
</blockquote></details>

Thanks for the review :)  

Done.

PR #303, opened 2018-08-27
Remove cruft from training_test.ts

caisq:
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs-layers/303)


caisq:
Thanks for catching that, @bileschi !

bileschi:
:lgtm_strong:

PR #305, opened 2018-08-28
Adds support for tensor.dot() for full set of compatible shapes

bileschi:
#### Description
FEATURE: Adds support for tensor dot product for additional tensor shapes, including x of dimension 4+ and y of dimension 3+.  Follows the API of python Keras.
---

---
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs-layers/305)

### DIFF
--- a/integration_tests/tfjs2keras/package.json
+++ b/integration_tests/tfjs2keras/package.json
@@ -4,7 +4,7 @@
   "description": "Testing sending models from tfjs-layers to Keras",
   "private": false,
   "dependencies": {
-    "@tensorflow/tfjs-core": "~0.12.14",
+    "@tensorflow/tfjs-core": "~0.12.15",
     "@tensorflow/tfjs-layers": "*",
     "@tensorflow/tfjs-node": "~0.1.13",
     "clang-format": "~1.2.2",
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@tensorflow/tfjs-layers",
-  "version": "0.7.4",
+  "version": "0.7.5",
   "description": "TensorFlow layers API in JavaScript",
   "private": false,
   "main": "dist/index.js",
@@ -10,7 +10,7 @@
   "jsdelivr": "dist/tf-layers.min.js",
   "unpkg": "dist/tf-layers.min.js",
   "devDependencies": {
-    "@tensorflow/tfjs-core": "~0.12.14",
+    "@tensorflow/tfjs-core": "~0.12.15",
     "@types/jasmine": "~2.5.53",
     "clang-format": "~1.2.2",
     "http-server": "~0.10.0",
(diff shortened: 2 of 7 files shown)


bileschi:
Fixes https://github.com/tensorflow/tfjs/issues/632

bileschi:
___
*[src/backend/tfjs_backend.ts, line 407 at r1](https://reviewable.io/reviews/tensorflow/tfjs-layers/305#-LL5aVhB7wTKxOgtKO6s:-LL5fTJo1AbHtnpnSdj-:b-89q77x) ([raw file](https://github.com/tensorflow/tfjs-layers/blob/b07a4e0a62743d21ae5b96f09b5de7a6d3da6a2e/src/backend/tfjs_backend.ts#L407)):*
<details><summary><i>Previously, dsmilkov (Daniel Smilkov) wrote…</i></summary><blockquote>

simpler to just do `x.reshape([-1, xLastDim])` and the library will figure out what the -1 is. Avoids the call to `.reduce((x, y) => x*y)`
</blockquote></details>

Cool!

bileschi:
___
*[src/backend/tfjs_backend.ts, line 387 at r2](https://reviewable.io/reviews/tensorflow/tfjs-layers/305#-LL5k6459xAFdG0sbNj4:-LL5kdqt2XG25Mvg-bYt:b5fcz28) ([raw file](https://github.com/tensorflow/tfjs-layers/blob/b68d3d288d6f2c4bb1a753f993c940c096f44e72/src/backend/tfjs_backend.ts#L387)):*

> ```
>  >= 3
> ```

I may be missing something here, but is this supposed to be `>= 2`?
</blockquote></details>

The code is correct whether we use >=3 or >=2 here.

If we keep >=3, we are treating these higher rank tensors as special case objects which undergo special scrutiny, and will get an error at this level if the shapes don't match.   If there is a shape mismatch and y is rank 2, the error raised will be the standard error from tf.dot

If we switch to >= 2 (which really could be no check at all, since rank <2 is caught above), then we are treating rank 2 as a special case of the general pattern, where the second last dim of y must match the last dim of x.   It's a matter of semantics.   

To me, the general case should be the standard 2d x 2d matmul, and any higher rank case should be spelled out here.

bileschi:
___
*[src/backend/tfjs_backend.ts, line 405 at r2](https://reviewable.io/reviews/tensorflow/tfjs-layers/305#-LL5iX2x9ifpu-b6zKD3:-LL5lljDEg8R-AO98WQq:b-b18wqp) ([raw file](https://github.com/tensorflow/tfjs-layers/blob/b68d3d288d6f2c4bb1a753f993c940c096f44e72/src/backend/tfjs_backend.ts#L405)):*

> ```
> xShape[xShape.length - 1];
> ```

Can `pop()` be used here for simplicity and consistency with below? 
</blockquote></details>

sure, done :)

PR #315, opened 2018-09-06
Fix a bug in the handling of array of step outputs by TimeDistributed layer

caisq:
Fixes: https://github.com/tensorflow/tfjs/issues/681

BUG

---
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs-layers/315)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -17,6 +17,6 @@ matrix:
       node_js: "8"
       stage: integ
       script:
-        - yarn test-integ
+        - echo 'yarn test-integ is disabled currently (https://github.com/tensorflow/tfjs/issues/679)'
       git:
         depth: 5
--- a/integration_tests/tfjs2keras/package.json
+++ b/integration_tests/tfjs2keras/package.json
@@ -4,9 +4,9 @@
   "description": "Testing sending models from tfjs-layers to Keras",
   "private": false,
   "dependencies": {
-    "@tensorflow/tfjs-core": "~0.12.16",
+    "@tensorflow/tfjs-core": "^0.12.18",
     "@tensorflow/tfjs-layers": "*",
-    "@tensorflow/tfjs-node": "~0.1.13",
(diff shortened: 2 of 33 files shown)


rodrigopivi:
hi @caisq

Thanks for this fix, after manually building the tfjs-layers and testing this use case of passing a model as a layer. This fix works when training with a `validationSplit` of `0.5`. If i use another validation split value, then the training and validation tensors will have different shapes and tfjs will throw an error. Here is an example code of this problem:

```javascript
const inputs = tf.input({ dtype: 'float32', shape: [1, 2] });
const lstm = tf.layers.lstm({ units: 2, returnSequences: true }).apply(inputs) as tf.SymbolicTensor;
const timeAttention = new TimeSeriesAttention({}).apply(lstm) as tf.SymbolicTensor;
const model = tf.model({ inputs, outputs: timeAttention });
const optimize = tf.train.adam(0.0066, 0.0025, 0.1);
model.compile({ loss: 'categoricalCrossentropy', metrics: ['accuracy'], optimizer: optimize });

const inp = tf.tensor3d([[[1,1]],[[2,2]],[[3,3]],[[4,4]],[[5,5]],[[6,6]]],[6,1,2]);
const out = tf.tensor3d([[[1,0]],[[2,0]],[[3,0]],[[4,0]],[[5,0]],[[6,0]]], [6, 1, 2]);

(async () => {
    // NOTE: if validation split is 0.5 this works, else it fails because the tensors shape for
    //             train and validation will have different shapes
    await model.fit(inp, out, { validationSplit: 0.2 });
})();
```

caisq:
@rodrigopivi Thanks for the report and the code for reproducing the error. It's on my TODO list to look at this issue. But my schedule is a little tight curerntly, so expect a delay of 1-2 weeks.

rodrigopivi:
thank you

PR #318, opened 2018-09-09
Use backward layer for computing yRev even when an initial state is given.

adarob:
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs-layers/318)


bileschi:
Thanks @adarob , it would have been nice if this bug had been exposed in a test.  Would it be possible / difficult to add a test that would be broken but for your fix?

caisq:
@adarob, +1 to what @bileschi  if you are willing to add the test that would be great. We'll also be happy to merge this PR first and add a test ourselves. Just let us know. Thanks.

caisq:
@bileschi Given we are planning a new minor release soon, I'll merge this PR first and add tests for it later. How does that sound?

bileschi:
@caisq works for me.

caisq:
@bileschi Thanks.

PR #320, opened 2018-09-10
Metrics: Adds precision to the api

manrajgrover:
#### Description

This PR adds the precision metric to the API. Currently, it doesn't take [weights](https://www.tensorflow.org/api_docs/python/tf/metrics/precision) into account. `truePositives` and `falsePositives` are currently internal operations which can be exposed as well.

---

---
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs-layers/320)

### DIFF
--- a/src/exports_metrics.ts
+++ b/src/exports_metrics.ts
@@ -60,6 +60,17 @@ export function categoricalCrossentropy(yTrue: Tensor, yPred: Tensor): Tensor {
  * @doc {
  *   heading: 'Metrics',
  *   namespace: 'metrics',
+ *   useDocsFrom: 'precision'
+ * }
+ */
+export function precision(yTrue: Tensor, yPred: Tensor): Tensor {
+  return metrics.precision(yTrue, yPred);
+}
+
+/**
+ * @doc {
+ *   heading: 'Metrics',
+ *   namespace: 'metrics',
  *   useDocsFrom: 'cosineProximity'
  * }
  */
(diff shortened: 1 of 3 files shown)


manrajgrover:
> So yPred can take values that are not 0 or 1, such as 0.75? If so, where is the thresholding logic? The test cases in metrics_test.ts don't seem to cover the non-0-or-1 case.

Nope, there is no thresholding done. Input should be a one-hot encoded tensor of the categories. Documentation was wrong. Fixed it.

Made rest of the changes.

manrajgrover:
> On a related note: the Python TF cast the input tensor to bool type, which means any non-zero value becomes true. Your implementation is slightly different though. It seems to require the values to be only 0 and 1. I think that's fine for now, but it should be stated clearly in the doc string above.

Yup, agreed. Casting to `bool` makes non-zero values become true. I don't understand the usecase here though. Why would we want to accept non-zero values for precision and cast them to `bool`? I haven't come across such a usecase. I feel if labels (non-zero) are being passed directly, they shouldn't be converted to `bool` and should be handled the way sklearn does. Made the change though.

> Are these really one-hot encoding of categories? Aren't these supposed to be just 0-1 boolean labels for the examples? For instance, suppose you have three input examples, and their true labels are true, false, and true. Isn't yTrue supposed to be simply [1, 0, 1], which is not a one-hot encoding? A one -hot encoding would be something like:
> 
> [[0, 1],
> [1, 0],
> [0, 1]], which I don't think is what the equivalent function in Python TF (tf.metrics.precision) expects.

Agreed. It should not be one-hot encoding for binary labels.

Made the changes.

manrajgrover:
@caisq Yep. Will build upon this PR (making fewer review iterations).

PR #321, opened 2018-09-10
Some performance optimization for RNNs

caisq:
* Keep the per-step outputs and concatenate them as a return value, if and only if it is necessary to do so. To this end, add arg `needPerStepOutputs` to the helper method rnn().
* Add `tidy()` call around the step-function call in `rnn()`.

Related:
* Clean up the signature of the helper method `rnn()`.

Benchmarking results, using the lstm-text-generation example (https://github.com/tensorflow/tfjs-examples/tree/master/lstm-text-generation):
  * Inference:
    * 7% reduction in inference time at LSTM layer size 128
    * 4% reduction at size 256
  * Training: 
    * 13% reduction in training time at LSTM layer size 128
    * 0.5% reduction at size 256

PERF

---
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs-layers/321)

### DIFF
--- a/src/layers/recurrent.ts
+++ b/src/layers/recurrent.ts
@@ -186,7 +186,7 @@ export function rnn(
   for (let t = 0; t < timeSteps; ++t) {
     let currentInput = K.sliceAlongFirstAxis(inputs, t, 1);
     currentInput = currentInput.reshape(currentInput.shape.slice(1));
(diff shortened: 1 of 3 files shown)


caisq:
cc @adarob

PR #332, opened 2018-10-04
Let Sequential.fromConfig() handle non-array config objects

caisq:
Previously, it was assumed that the config object passed to a
Sequential model during serialization is always an array (of
layer configs). But in TensorFlow r1.11, a change was made to
the serialization of Sequential, such that the config is now
an object with the 'layers' field.

This PR makes sure that TensorFlow.js is able to load such
serialized Sequential models without error.

BUG

Fixes: https://github.com/tensorflow/tfjs/issues/744

---
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs-layers/332)

### DIFF
--- a/src/models.ts
+++ b/src/models.ts
@@ -822,7 +822,7 @@ export class Sequential extends Model {
       cls: serialization.SerializableConstructor<T>,
       config: serialization.ConfigDict): T {
     let configArray: serialization.ConfigDictArray;
-    let additionalModelConfig: serialization.ConfigDict = {};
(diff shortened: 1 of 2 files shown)


cbasavaraj:
Hi, I still get an error when on tf.loadModel(MODEL_URL). 

> Unhandled Promise Rejection: Error: Sequential.fromConfig called without an array of configs

I have upgraded to the latest versions of Keras and tfjs. Am I missing anything? Thanks

PR #343, opened 2018-10-19
Fix deserialization of GlorotUniform & others; Fix tensor name regex

caisq:
BUG

Fixes https://github.com/tensorflow/tfjs/issues/798

---
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs-layers/343)


caisq:
Thanks for the review, @bileschi !

PR #344, opened 2018-10-21
Expose modelFromJSON as tf.models.modelFromJSON

caisq:
Fixes https://github.com/tensorflow/tfjs/issues/815

This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs-layers/344)


src/models.ts
```diff
@@ -38,6 +38,9 @@ import {getExactlyOneShape} from './utils/types_utils';
  *       considered during deserialization.
  * @returns A TensorFlow.js Layers `Model` instance (uncompiled).
  */
+/**
+ * @doc {heading: 'Models',subheading: 'Loading'}
+ */
```
bileschi: The other two api items under this heading include descriptions and sample code.  Does it make sense to include at least the simplest sample code here?

something like:

```
myJsonString = ...;
const model = await tf.models.modelFromJSON(myJsonString);
model.summary();
```

PR #346, opened 2018-10-22
Fix issues in loading an attention model

caisq:
BUG

* Fix an incorrect query into a Set object in container.ts. Previously the `in`
  keyword was used, which didn't actually work (it prevented loading
  functional models with complex topology from working properly)
* Fix a bug that caused an infinite loop in container.ts.
* Fix a bug wherein the `Dot.computeOutputShape()` method altered input
  tensor shape inadvertently.
* Add unit tests for the aforementioned fixes.

Fixes: https://github.com/tensorflow/tfjs/issues/794

---
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs-layers/346)


caisq:
Thanks for the reviews!

PR #360, opened 2018-11-01
Improve integration_test/benchmarks

caisq:
- Add two new benchmarks: MobileNet and an Attention-based translation model
- Allow some benchmarks to run inference only (no training)
- Use `yarn watch` instead of `http-server` from the main script (build-benchmarks.sh)
- Dispose old inputs/targets better in index.js

Relating to https://github.com/tensorflow/tfjs/issues/863

This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs-layers/360)


integration_tests/benchmarks/build-benchmarks.sh
```diff
@@ -48,15 +48,16 @@ if [[ "${SKIP_PY_BENCHMAKRS}" == 0 ]]; then
...
-echo
-
-node_modules/http-server/bin/http-server -p "${DEMO_PORT}" dist/
+echo "Building benchmarks demo..."
+cd ${DEMO_DIR}
+# Force a refresh of the tfjs-layers dependency.
+rm -rf node_modules
+yarn
+yarn build
+yarn watch
```
ericdnielsen: I think we want a banner again of some type (like the removed code) saying:
'Press the 'run benchmakrs' button in the launched browser to execute the TFJS Benchmarks.

integration_tests/benchmarks/index.js
```diff
@@ -28,32 +28,65 @@ async function runBenchmark(artifactsDir, modelName, config) {
   // for losses.
 
   const batchSize = benchmarkData.batch_size;
-  const xs = tfc.randomUniform([batchSize].concat(benchmarkData.input_shape));
-  const ys = tfc.randomUniform([batchSize].concat(benchmarkData.target_shape));
-  model.compile({
-    optimizer: benchmarkData.optimizer,
-    loss: lossMap[benchmarkData.loss],
-  });
+  let xs;
```
ericdnielsen: Can we pull the shape inference out into a helper function?  (Yes I doubt it would be reused, but this runBenchmark function is getting long/harder to follow.

integration_tests/benchmarks/python/benchmarks.py
```diff
@@ -69,17 +69,40 @@ def benchmark_and_serialize_model(model_name,
     2. Average predict() time over all the _PREDICT_RUNS.
   """
   model = model_fn(input_shape, target_shape)
-  model.compile(optimizer=optimizer, loss=loss)
-  xs = np.random.rand(*([batch_size] + input_shape))
-  ys = np.random.rand(*([batch_size] + target_shape))
+  if train_epochs:
+    model.compile(optimizer=optimizer, loss=loss)
+
+  if input_shape is None:
```
ericdnielsen: Same comment as ts code regarding pulling this code out to a helper.

integration_tests/benchmarks/build-benchmarks.sh
caisq: Done.

PR #364, opened 2018-11-04
Always use implementation 2 in GRU and LSTM for performance

caisq:
PERF

Also in this change:
- Fix a small bug in the logic of GRU dropout
- Add more unit tests to ensure faithful preservation of `implementation` in the config for GRU and LSTM, even though it is ignored in terms of actual implementation in TensorFlow.js

Benchmark results:
- From integration_tests/benchmark:
    - GRU
      - predict: 106 ms --> 83.2 ms (1.27x)
      - fit: 180 ms --> 106 ms (1.7x)
    - LSTM
      - predict: 137.7 ms --> 64.3 ms (2.1x)
      - fit: 224.0 ms --> 100 ms (2.24x)
    - Attention
      - predict: 148.9 ms --> 77.9 ms (1.9x)

Similar speed-up is observed for tfjs-node:
- LSTM input shape: [null, 100, 128]:  predict:1226.7 ms --> 602.2 ms (2.0x)

Relating to https://github.com/tensorflow/tfjs/issues/863

This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs-layers/364)

### DIFF
--- a/src/layers/recurrent_test.ts
+++ b/src/layers/recurrent_test.ts
@@ -1434,14 +1434,16 @@ describeMathCPU('LSTM Symbolic', () => {
      });
 
 
-  it('Serialization round trip', () => {
-    const layer = tfl.layers.lstm({units: 4});
-    const pythonicConfig = convertTsToPythonic(layer.getConfig());
-    // tslint:disable-next-line:no-any
-    const tsConfig = convertPythonicToTs(pythonicConfig) as any;
-    const layerPrime = tfl.layers.lstm(tsConfig);
-    expect(layerPrime.getConfig().units).toEqual(4);
-  });
+  for (const implementation of [1, 2]) {
+    it('Serialization round trip', () => {
+      const layer = tfl.layers.lstm({units: 4, implementation});
+      const pythonicConfig = convertTsToPythonic(layer.getConfig());
+      // tslint:disable-next-line:no-any
+      const tsConfig = convertPythonicToTs(pythonicConfig) as any;
+      const layerPrime = tfl.layers.lstm(tsConfig);
+      expect(layerPrime.getConfig().units).toEqual(4);
+    });
+  }
 });
 
 describeMathCPUAndGPU('LSTM Tensor', () => {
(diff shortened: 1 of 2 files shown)


src/layers/recurrent_test.ts
```diff
@@ -970,14 +970,16 @@ describeMathCPU('GRU Symbolic', () => {
...
-    expect(layerPrime.getConfig().units).toEqual(4);
-  });
+  for (const implementation of [1, 2]) {
+    it('Serialization round trip', () => {
+      const layer = tfl.layers.gru({units: 4, implementation});
+      const pythonicConfig = convertTsToPythonic(layer.getConfig());
+      // tslint:disable-next-line:no-any
+      const tsConfig = convertPythonicToTs(pythonicConfig) as any;
+      const layerPrime = tfl.layers.gru(tsConfig);
+      expect(layerPrime.getConfig().units).toEqual(4);
```
ericdnielsen: do we need an assert/expect that the implementation value was preserved?

src/layers/recurrent_test.ts
```diff
@@ -1432,14 +1434,16 @@ describeMathCPU('LSTM Symbolic', () => {
...
-    expect(layerPrime.getConfig().units).toEqual(4);
-  });
+  for (const implementation of [1, 2]) {
+    it('Serialization round trip', () => {
+      const layer = tfl.layers.lstm({units: 4, implementation});
+      const pythonicConfig = convertTsToPythonic(layer.getConfig());
+      // tslint:disable-next-line:no-any
+      const tsConfig = convertPythonicToTs(pythonicConfig) as any;
+      const layerPrime = tfl.layers.lstm(tsConfig);
+      expect(layerPrime.getConfig().units).toEqual(4);
```
ericdnielsen: do we need an assert/expect that the implementation value was preserved?

src/layers/recurrent_test.ts
caisq: Thanks for the review, @ericdnielsen !

Good catch! Done.

src/layers/recurrent_test.ts
caisq: Good catch! Done.

PR #388, opened 2018-12-11
Add general masking support

caisq:
- Executor now understands and executes masks
- Add unit tests

Some layers still do not have detailed logic for masking, including:
- Bidrectional
- TimeDistributed

FEATURE

This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs-layers/388)

### DIFF
--- a/src/engine/container.ts
+++ b/src/engine/container.ts
@@ -842,6 +842,7 @@ export abstract class Container extends Layer {
         masks = generic_utils.toList(mask);
       }
       // TODO(michaelterry): Add support for mask caching.
+      // TODO(cais): Get rid of runInternalGraphs(). DO NOT SUBMIT.
       return this.runInternalGraph(inputs, masks)[1];
     });
   }
--- a/src/engine/container.ts
+++ b/src/engine/container.ts
@@ -833,6 +833,7 @@ export abstract class Container extends Layer {
    */
   computeMask(inputs: Tensor|Tensor[], mask?: Tensor|Tensor[]): Tensor
       |Tensor[] {
+    console.log('**** In Container.computeMask()');  // DEBUG
     return tidy(() => {
       inputs = generic_utils.toList(inputs);
       let masks: Tensor[];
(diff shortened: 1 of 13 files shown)


bileschi:
Should I look to a python implementation to compare this to?  I know that keras-team/keras has been significantly refactored since tfjs-layers was built and there is no more executor so I wonder should I look to an older commit, or is most of this code brand new for tfjs?

bileschi:
___
*[src/engine/container.ts, line 845 at r1](https://reviewable.io/reviews/tensorflow/tfjs-layers/388#-LTUc7Yx4vX_jKj-FkPT:-LTUc7Yx4vX_jKj-FkPU:bxd9f5d) ([raw file](https://github.com/tensorflow/tfjs-layers/blob/fd325a4d557b98ebc7574ea677c4428465535e38/src/engine/container.ts#L845)):*
> ```
> // TODO(cais): Get rid of runInternalGraphs(). DO NOT SUBMIT.
> ```
This seems ominous...

bileschi:
___
*[src/layers/embeddings.ts, line 130 at r1](https://reviewable.io/reviews/tensorflow/tfjs-layers/388#-LTUeHkf-VQ-xkfNICmb:-LTUeHkf-VQ-xkfNICmc:buikt38) ([raw file](https://github.com/tensorflow/tfjs-layers/blob/fd325a4d557b98ebc7574ea677c4428465535e38/src/layers/embeddings.ts#L130)):*
> ```
> this.supportsMasking = config.maskZero;
> ```
This seems strange, but I see that it's the same as the python implementation.  Why wouldn't the embedding layer always "supportMasking".  Why are there two configurations that are the same thing here?

bileschi:
___
*[src/engine/executor.ts, line 87 at r1](https://reviewable.io/reviews/tensorflow/tfjs-layers/388#-LTUfOhV9ZDmf4f8eaOf:-LTUfOhV9ZDmf4f8eaOg:b-b1d6qj) ([raw file](https://github.com/tensorflow/tfjs-layers/blob/fd325a4d557b98ebc7574ea677c4428465535e38/src/engine/executor.ts#L87)):*
> ```XML
>    * Add a key-value pair to the FeedDict.
>    * @param key The key of the feed.
>    * @param value The value of the feed.
> ```

nit: missing new mask param in JSDoc here.

PR #391, opened 2018-12-13
Add unit test for RNN with initialState input

caisq:
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs-layers/391)


caisq:
Thanks, @bileschi !

PR #397, opened 2019-01-03
Call fused matMul in core whenever possible for performance.

annxingyuan:
This PR addresses https://github.com/tensorflow/tfjs/issues/825

#### Benchmarks

Test case: Sequential model with input shape `[100, 100, 1]` and 10 dense layers, each with bias and relu activation. 

Average prediction time over 100 runs:

`master`: 13.5ms

`fuse_activations`: 10.9ms

#### Description
- Calls new `fused.matMul` in tfjs-core when a supported activation kernel is used for a `Dense` layer

---

---
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs-layers/397)


caisq:
Thanks for the PR, @annxingyuan Two questions before I review the PR

* It looks like there are some test failures due to `tf.fused` being missing in the version of -core that -layers depends on. Any thoughts on how to fix it? Does it require a new release of -core?
* Do you have any benchmarks on how much this improves model performance?

annxingyuan:
@caisq 

It looks like they are - I'll fix them and ping you once everything passes! BTW I added some benchmarks in the description.

annxingyuan:
@caisq Hey Shanqing - could you take another look at this PR when you get a chance? Tests are passing now - there was a fix in core 1.0 for the previous issue.

annxingyuan:
Thanks @caisq - could you take another look? I added benchmarking results to the PR description.

PR #398, opened 2019-01-06
Model.fit(): Push await data() on scalar log values more toward the end 

caisq:
PERF

Benchmark changes:

Number of benchmark runs: 1 per condition

Model.fit() time:
MODEL_NAME: BEFORE --> AFTER (time unit: ms) (percent change)
dense-tiny: 15.0 --> 10.2 (-32%)
dense-large: 74.1 --> 64.6 (-12.8%)
convolutional-1filters: 78.1 --> 62.5 (-20.0%)
convolutional-32filters: 567.7 --> 561.0 (-1.2%)
rnn-simpleRNN: 31.9 --> 31.4 (-1.6%)
rnn-GRU: 89.2 --> 89.2 (0%)
rnn-LSTM: 74.5 --> 69.7 (-6.4%)

Fixes: https://github.com/tensorflow/tfjs/issues/1066

---
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs-layers/398)


caisq:
cc @annxingyuan

dsmilkov:
Yes, correct. await data() is at least 5ms yield. nextFrame() adds another 5 ms, so calling both will result in `>=10ms` cpu rest.

PR #402, opened 2019-01-07
Fix a memory leak in addWeight()

caisq:
The intial-value tensor was not properly disposed, but is now.

BUG

Fixes: https://github.com/tensorflow/tfjs/issues/995

---
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs-layers/402)


caisq:
Thanks for the review, @nsthorat!

PR #403, opened 2019-01-07
Rename JsonDict to PyJsonDict for clarity

davidsoergel:
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs-layers/403)

### DIFF
--- a/src/types.ts
+++ b/src/types.ts
@@ -49,18 +49,40 @@ export type NamedTensorMap = {
 };
 
 /**
- * Types to support JSON.
+ * A value within the JSON-serialized form of a serializable object.
+ *
+ * The keys of any nested dicts should be in snake_case (i.e., using Python
+ * naming conventions) for compatibility with Python Keras.
+ *
+ * @see PyJsonDict
+ */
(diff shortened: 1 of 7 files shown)


src/types.ts
```diff
@@ -57,16 +57,16 @@ export type NamedTensorMap = {
  * converted to a ConfigDict that has renamed fields (TS naming conventions)
  * and support for Enums.
  */
-export type JsonValue = boolean|number|string|null|JsonArray|JsonDict;
-export interface JsonDict {
-  [key: string]: JsonValue;
+export type PyJsonValue = boolean|number|string|null|PyJsonArray|PyJsonDict;
```
caisq: Can you add a sentence or two to the doc string of this type to describe why it's called `PyJsonValue` (i.e., that it contains keys in Python Keras-compatible format, that is, snake_case)

src/types.ts
```diff
@@ -57,16 +57,16 @@ export type NamedTensorMap = {
...
+export interface PyJsonDict {
```
caisq: Same here. Please add a comment explaining what the "Py" part of the new interface name means.

PR #404, opened 2019-01-07
 Rename *Config to *Args throughout

davidsoergel:
First in a chain of PRs to make the distinction between the serialized Keras JSON format and the tfjs-layers constructor arguments.

BREAKING

This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs-layers/404)

### DIFF
--- a/src/base_callbacks.ts
+++ b/src/base_callbacks.ts
@@ -179,7 +179,6 @@ export class CallbackList {
     if (logs == null) {
       logs = {};
     }
-    await resolveScalarsInLogs(logs);
     for (const callback of this.callbacks) {
       await callback.onBatchEnd(batch, logs);
     }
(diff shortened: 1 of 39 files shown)


bileschi:
___
*[src/layers/convolutional.ts, line 1025 at r1](https://reviewable.io/reviews/tensorflow/tfjs-layers/404#-LVfM2nt75ty-mgcYPfD:-LViCfHj82A-bifipMHh:b-qwtopa) ([raw file](https://github.com/tensorflow/tfjs-layers/blob/7ba75ca501f536b5943fe01be1fe0ce0883ecfed/src/layers/convolutional.ts#L1025)):*

Oh, I just noticed that it's not used outside of this class hierarchy so it didn't need to be public.  This is an unrelated cleanup; happy to revert if anyone cares.
</blockquote></details>

No need to revert so long as it was intentional

PR #407, opened 2019-01-08
Move Shape and DataType into src/keras_format (3/n)

davidsoergel:
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs-layers/407)


davidsoergel:

PR #408, opened 2019-01-08
Clarify which *Args are actually *NonSerializableArgs (4/n)

davidsoergel:
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs-layers/408)

### DIFF
--- a/src/base_callbacks.ts
+++ b/src/base_callbacks.ts
@@ -469,7 +469,7 @@ export class History extends BaseCallback {
   }
 }
 
-export interface CustomCallbackArgs {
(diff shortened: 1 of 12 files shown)


ericdnielsen:
I'm not sure I see the point of this PR in the chain.  Isn't the end state such that all Args are nonSerializable, that's Config's job?  So why inflate the name of everything with a long tag?

ericdnielsen:
So I think that means 406 and onward will need tweaks to remove NonSerializable, right?

davidsoergel:

PR #409, opened 2019-01-08
 Clarify that Layer Args are not serializable (because weights) (5/n)

davidsoergel:
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs-layers/409)


davidsoergel:

PR #410, opened 2019-01-08
Factor out core layer configs. (6/n)

davidsoergel:
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs-layers/410)


davidsoergel:

PR #412, opened 2019-01-08
Move *Identifier into src/keras_format (8/n)

davidsoergel:
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs-layers/412)


davidsoergel:

PR #415, opened 2019-01-09
Move Shape and DataType into src/keras_format (2/n)

davidsoergel:
FEATURE Provide types describing the Keras JSON format (2/n)

---
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs-layers/415)


davidsoergel:
This is just like #407 (which was already approved)-- just recreated for a cleaner git history.  Submitting on the basis of the prior approval.

PR #416, opened 2019-01-10
Move ActivationIdentifier into src/keras_format (3/n)

davidsoergel:
This is very confusing!  Activation functions are serialized only as a string identifier, not as a `*Serializable` type.  The string identifier is given by a `static readonly className` field in each class, which may differ from the actual class name (in capitalization, at least).  The identifiers are lowerCamelCase.  The `ActivationIdentifier` type recapitulates exactly the strings given in the `className` fields.  Hence, the `ActivationIdentifier` type is needed as part of the serialization format spec.

That is all in contrast to Constraints, Initializers, and Regularizers.  In those cases, serialization is done via a full `*Serializable` type, where the `class_name` field is populated from the actual class name (always UpperCamelCase).  In those cases, the values of `ConstraintIdentifier`, `InitializerIdentifier`, and `RegularizerIdentifier` are in lowerCamelCase, and they are used only in the course of accepting constructor arguments during programmatic use of the API.  Thus, those `*Identifier` types *do not* form part of the serialization spec.

FEATURE Provide types describing the Keras JSON format (3/n)

---
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs-layers/416)

### DIFF
--- a/src/activations.ts
+++ b/src/activations.ts
@@ -12,8 +12,9 @@
 import * as tfc from '@tensorflow/tfjs-core';
 import {serialization, Tensor, tidy} from '@tensorflow/tfjs-core';
 
-import * as K from './backend/tfjs_backend';
 import {getScalar} from './backend/state';
+import * as K from './backend/tfjs_backend';
+import {ActivationIdentifier} from './keras_format/activation_config';
 import {deserializeKerasObject} from './utils/generic_utils';
 
 /**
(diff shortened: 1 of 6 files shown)


davidsoergel:
@ericdnielsen looking forward to your hard-won expertise on this one :)

davidsoergel:
Slight correction: for Constraints, Initializers, and Regularizers, the serialized `class_name` is also taken from a `static readonly className` field-- it just happens that the values are UpperCamelCase and match the actual class name.

Finally: I think the serialized `class_name` fields for the anomalous VarianceScaling subclasses (GlorotUniform etc.) should have been UpperCamelCase, not snake_case, so I fixed those in this PR also (since this is all part of the same cluster of confusions).

davidsoergel:
The integration test failure is unrelated (looks like a tfjs-core version mismatch issue?), so submitting anyway.

caisq:
@davidsoergel Yep, it's unrelated. It's due to some out of sync ops between the new version of tfjs union and tfjs-node. Nick has a PR that'll fix it soon.

PR #417, opened 2019-01-10
 Provide core layer configs in src/keras_format (4/n)

davidsoergel:
FEATURE Provide types describing the Keras JSON format (4/n)

---
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs-layers/417)


davidsoergel:
I felt that the GitHub chained-PR approach was too confusing before, so I'm going back to 'Staged' PRs against master.  These accumulate changes from all prior outstanding PRs.  I want to establish the PRs so you can see what's next, but I won't actually request review until all of the upstream ones are merged.

PR #419, opened 2019-01-10
Add all the rest of the layer serializations

davidsoergel:
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs-layers/419)


caisq:
Fine to punt naming. Your point of leaving out "_serialization" causing confusion is a good one. How about naming them like "advanced_activations_specs.ts"? It's
- more to the point
- no risk of confusion
- shorter

PR #420, opened 2019-01-11
Enforce *Serialization is JSON-like. Add top level, training_config etc. (7/n)

davidsoergel:
This PR does three things (sorry):

1. Enforce that all *Serialization types extend PyJsonDict, so the types are definitely and easily serializable.
2. Add the training_config stuff in addition to model_config.
3. Add the top-level structure, so that the complete Keras JSON file has a type.

This did require creating *Config types throughout after all, instead of inlining them in *Serialization.

---
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs-layers/420)

### DIFF
--- a/src/keras_format/types.ts
+++ b/src/keras_format/types.ts
@@ -44,3 +44,18 @@ export interface PyJsonDict {
  * @see PyJsonDict
  */
 export interface PyJsonArray extends Array<PyJsonValue> {}
+
+/**
+ * A Keras JSON entry representing a Keras object such as a Layer.
+ *
+ * The Keras JSON convention is to provide the `class_name` (e.g., the layer
+ * type) at the top level, and then to place the class-specific configuration in
+ * a `config` subtree.  These class-specific configurations are provided by
+ * subtypes of `PyJsonDict`.  Thus, this `*Serialization` has a type parameter
+ * giving the specific type of the wrapped `PyJsonDict`.
+ */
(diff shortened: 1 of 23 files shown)


src/keras_format/types.ts
```diff
@@ -44,3 +44,18 @@ export interface PyJsonDict {
...
+/**
+ * A Keras JSON entry representing a Keras object such as a Layer.
+ *
+ * The Keras JSON convention is to provide the `class_name` (e.g., the layer
+ * type) at the top level, and then to place the class-specific configuration in
+ * a `config` subtree.  These class-specific configurations are provided by
+ * subtypes of `PyJsonDict`.  Thus, this `*Serialization` has a type parameter
+ * giving the specific type of the wrapped `PyJsonDict`.
+ */
+export interface BaseSerialization<N extends string, T extends PyJsonDict>
```
ericdnielsen: Pretty sure what I'm asking for is impossible, but wanted to at least mention -- all the types that extend this look an awful lot like the register for serializablility code.  Of course this is spec side, not implementation side.  But would be nice if there were a way to automatically? register classes that extend this.

bileschi:
*[a discussion](https://reviewable.io/reviews/tensorflow/tfjs-layers/420#-LVxbUu3CCYOTdoEdCMN:-LVxbUu3CCYOTdoEdCMO:b-tw6r4p) (no related file):*
Is this code exercised somewhere?  Is there some sort of testing that should be done here?

---

PR #433, opened 2019-01-23
Expose valid identifiers and class names for various Keras categories

davidsoergel:
In https://github.com/tensorflow/tfjs-layers/pull/423, we used a class to represent a map from valid Loss identifiers to human-friendly names for those losses.  From that class, we also generated a union type of the loss identifiers.

Here, we extend that approach to Activations, Constraints, Initializers, Optimizers, and Regularizers.

For the latter four cases, the options are not merely string identifiers but rather nested object instances (represented by *Serialization).  In those cases, we derive a runtime-available `string[]` of the valid class names from the existing type-level information.  Finally, we test that the compile-time and runtime representations of the various options are in sync.

---
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs-layers/433)


src/keras_format/activation_config.ts
```diff
@@ -8,6 +8,38 @@
...
-/** @docinline */
-export type ActivationIdentifier = 'elu'|'hardSigmoid'|'linear'|'relu'|'relu6'|
-    'selu'|'sigmoid'|'softmax'|'softplus'|'softsign'|'tanh'|string;
+import {stringDictToArray} from './utils';
+
+/**
+ * List of all known activation names, along with a string description.
+ *
+ * Representing this as a class allows both type-checking using the keys and
+ * generating an appropriate options array for use in select fields.
```
ericdnielsen: This feels a little odd in the spec/keras_format.  We're coupling a particular apps's usage of fields into the spec?

src/keras_format/activation_config.ts
```diff
@@ -8,6 +8,38 @@
...
+  public readonly softplus = 'Softplus';
+  public readonly softsign = 'Softsign';
+  public readonly tanh = 'tanh';
+}
+
+/**
+ * An array of `{value, label}` pairs describing the valid activations.
+ *
+ * The `value` is the serializable string constant, and the `label` is a more
+ * user-friendly description (e.g. for use in UIs).
```
ericdnielsen: ditto.

PR #434, opened 2019-01-25
Change spies to conform with window.fetch API.

annxingyuan:
This PR addresses https://github.com/tensorflow/tfjs/issues/1140

#### Changes
- Return `Response` within a promise in stubbed fetch so it's `catch`-able
- Send content type header with fetch

---
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs-layers/434)

### DIFF
--- a/src/models_test.ts
+++ b/src/models_test.ts
@@ -8,19 +8,21 @@
  * =============================================================================
  */
 
-import {DataType, io, ones, randomNormal, Scalar, scalar, serialization, sum, Tensor, tensor1d, tensor2d, zeros, tensor3d} from '@tensorflow/tfjs-core';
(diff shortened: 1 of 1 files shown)


src/models_test.ts
```diff
@@ -546,7 +549,11 @@ describeMathCPU('loadModel from URL', () => {
       (fileBufferMap:
            {[filename: string]: Float32Array|Int32Array|ArrayBuffer}) => {
         spyOn(window, 'fetch').and.callFake((path: string) => {
-          return new Response(fileBufferMap[path]);
+          return new Promise((res, rej) => {
```
caisq: Google style prefers explicit, unambiguous names to short, succinct ones. So use `resolve` and `reject` instead.

caisq: Does it make sense to remove the curly brackets in this arrow function here, since it contains only one statement?

annxingyuan: Good point - I've removed them.

annxingyuan: Thanks - I've updated the names.

PR #442, opened 2019-01-30
Make tfjs-layers compatible with the closure compiler

dsmilkov:
- Add `/** @nocollapse */` on static properties to tell closure not to remove them.
- Prepend `declare` in front of interfaces and types that are external (used in json deserialization/serialization) so closure doesn't rename their properties.
- Instead of object literals, explicitly assign each field separately to a config object and mark it as `serialization.ConfigDict` so closure doesn't rename its properties.

Similar changes in tfjs-core: https://github.com/tensorflow/tfjs-core/pull/1521

---
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs-layers/442)

### DIFF
--- a/src/backend/state.ts
+++ b/src/backend/state.ts
@@ -40,10 +40,7 @@ export function getUid(prefix = ''): string {
   return prefix + _uidPrefixes[prefix].toString();
 }
 
-const scalarCache: {[typeKey: string]: {[key: number]: Scalar}} = {
-  float32: {},
-  int32: {}
-};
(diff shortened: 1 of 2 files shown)


caisq:
cc @ericdnielsen , @bileschi , @davidsoergel as an FYI.

caisq:
@dsmilkov Thank you for spending the time and effort in making the code compile with closure. 

Is there a way to ensure that going forward, people making changes in tfjs-layers will not break the closure compiler?

PR #449, opened 2019-02-05
remove the accept header check

pyu10055:
tfjs-core removed accept header for http model loader, updating the test here accordingly.

---
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs-layers/449)

### DIFF
--- a/src/models_test.ts
+++ b/src/models_test.ts
@@ -878,27 +878,12 @@ describeMathCPU('loadModel from URL', () => {
        expect(requestHeaders[0]).toEqual(jasmine.objectContaining({
          'header_key_1': 'header_value_1'
        }));
-       if (requestHeaders[0]['Accept']) {
-         expect(requestHeaders[0]).toEqual(jasmine.objectContaining({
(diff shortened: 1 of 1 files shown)


nsthorat:
Can we do this the same way we did the tfjs-converter by mocking the call to the core method?

src/models_test.ts
```diff
@@ -878,27 +878,12 @@ describeMathCPU('loadModel from URL', () => {
...
-       if (requestHeaders[0]['Accept']) {
-         expect(requestHeaders[0]).toEqual(jasmine.objectContaining({
-           'Accept': 'application/json'
-         }));
-       }
        expect(requestHeaders[1]).toEqual(jasmine.objectContaining({
          'header_key_1': 'header_value_1'
        }));
-       if (requestHeaders[1]['Accept']) {
-         expect(requestHeaders[1]).toEqual(jasmine.objectContaining({
```
nsthorat: in the future let's avoid if statements like this in unit tests

PR #457, opened 2019-02-11
Add assertions for positive-integer values

caisq:
FEATURE
UX

Fixes https://github.com/tensorflow/tfjs/issues/882

---
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs-layers/457)

### DIFF
--- a/src/layers/convolutional.ts
+++ b/src/layers/convolutional.ts
@@ -334,6 +334,7 @@ export abstract class BaseConv extends Layer {
     super(args as LayerArgs);
     BaseConv.verifyArgs(args);
     this.rank = rank;
+    generic_utils.assertPositiveInteger(this.rank, 'rank');
     if (this.rank !== 1 && this.rank !== 2) {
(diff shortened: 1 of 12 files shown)


nsthorat:
FYI I'm also doing this in core: https://github.com/tensorflow/tfjs-core/pull/1562

caisq:
@nsthorat Doing it in the core is useful. However, I think making assertions at the level of tfjs-layers still helps, because it leads to clearer error messages. WDYT?

nsthorat:
Oh yeah I wasn't saying ditch this PR -- this should absolutely happen here too!

caisq:
@nsthorat yeah, I considered using JSON.stringify(). The problem: it is expensive and sometimes leads to undesirable effects. For example, if someone accidentally passes a Model or Layer object as the value of units, it'll lead to a failure to convert to JSON or a huge error message containing a new JSON string.

PR #475, opened 2019-02-26
Switch all asserts to take a function that returns a string.

nsthorat:
This allows us to lazily evaluate the string for better performance.

Currently string concatenating a tensor happens eagerly, causing a dataSync() on the tensor and slows down the train loop.

---
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs-layers/475)

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -20,6 +20,7 @@
   "clang-format.style": "Google",
   "files.insertFinalNewline": true,
   "editor.detectIndentation": false,
+  "editor.rulers": [80],
   "editor.wrappingIndent": "none",
   "typescript.tsdk": "node_modules/typescript/lib",
   "typescript.format.insertSpaceAfterOpeningAndBeforeClosingNonemptyBraces": false,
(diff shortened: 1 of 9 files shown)


nsthorat:
Yes -- a major one :).

Each tick of fitDataset on a simple transfer learning webcam model on my workstation (which is a powerful 12 core machine) was taking 80ms for one assert (when each tick of the actual call to `minimize()` was taking 10ms) because of one of the asserts concatenating a string with a tensor, which calls toString(), which calls dataSync(). This means the tensor was uploading / downloading constantly and of course we had a sync.

So this should be a pretty major performance win across the board.

caisq:
@nsthorat Good to know! Looking forward to running the benchmarks and seeing the performance gains. It was my hunch that some major inefficiency like this existed in the code base across tfjs-core and tfjs-layer. Thanks for actually making the improvement. Anyway, I'll keep you posted about what I find in the layers benchmarks.

nsthorat:
@caisq it's possible you won't see the improvements, the change that added that `dataSync` was pretty recent: https://github.com/tensorflow/tfjs-layers/pull/458

PR #490, opened 2019-03-06
 Fix a few Camel- vs. Snake- case issues in keras_format.

davidsoergel:
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs-layers/490)

### DIFF
--- a/src/keras_format/activation_config.ts
+++ b/src/keras_format/activation_config.ts
@@ -21,4 +21,11 @@ export const activationOptions = stringLiteralArray([
 /**
  * A type representing the strings that are valid loss names.
  */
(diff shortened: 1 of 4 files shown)


bileschi:
Is this ready to submit?  Is there something that is blocking this?

nsthorat:
Is this a breaking change?

bileschi:
I don't think anything depends on keras_format yet, and the tests pass, so I'm inclined to volunteer that the answer is no.  @ericdnielsen do you know anything about this change?  Is it safe to submit?

ericdnielsen:
I'm pretty sure its safe/non-breaking.

PR #492, opened 2019-03-07
Allow `null`(`None` in Python) in `Shape`

huan:
See: https://github.com/tensorflow/tfjs/issues/1341

---
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs-layers/492)

### DIFF
--- a/src/engine/input_layer_test.ts
+++ b/src/engine/input_layer_test.ts
@@ -108,6 +108,13 @@ describeMathCPU('InputLayer', () => {
     });
   });
 
+  // See https://github.com/tensorflow/tfjs/issues/1341
+  it('allow `null` in shape', () => {
+    const inputShape = [null, 2];
+    const inputs = tfl.layers.inputLayer({inputShape});
+    expect(inputs.inputSpec[0].shape).toEqual([null].concat(inputShape));
+  });
+
   it('throws an exception if both inputShape and batchInputShape ' +
          'are specified during initialization.',
      () => {
--- a/src/keras_format/common.ts
+++ b/src/keras_format/common.ts
@@ -8,6 +8,7 @@
  * =============================================================================
  */
 
+// TODO(huan): add layer-specific input shape types (see: https://github.com/tensorflow/tfjs-layers/pull/492)
 /** @docalias (null | number)[] */
 export type Shape = (null | number)[];
 
--- a/src/keras_format/common.ts
+++ b/src/keras_format/common.ts
@@ -10,7 +10,7 @@
 
 // TODO(huan): add layer-specific input shape types (see: https://github.com/tensorflow/tfjs-layers/pull/492)
 /** @docalias (null | number)[] */
-export type Shape = (null | number)[];
+export type Shape = Array<null | number>;
 
 // The tfjs-core version of DataType must stay synced with this.
 export type DataType = 'float32'|'int32'|'bool'|'complex64'|'string';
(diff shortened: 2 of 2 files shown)


davidsoergel:
I agree with this in principle (mod the `strictNullChecks` point on the bug), but I don't have a good sense of whether it is safe.  It would require quite a lot of tests to prove that this works throughout TF.js.  Can you add at least a few tests, e.g. building a simple model with an unspecified batch size, and showing that it works for training and inference?

There are some edge cases where unspecified batch sizes definitely cause trouble, such as `tf.conv2DTranspose()` (IIRC).  That's also the case in Python Keras / TF so it's not a regression in that sense, but it still bears thinking about and testing.

It sounds like you effectively already showed that it works in your case by casting your shape to `number[]`-- that's promising.

@caisq, do you have an intuition about whether this is likely to just work throughout?

caisq:
@davidsoergel There are certainly many layer types that will not work right if certain dimensions of  the input shape are `null`. But there are also plenty of layer types that can handle `null` size in certain dimensions correctly. The time dimension of the inputs to RNNs is the most typical example.

As for typing, there are three approaches
1) The status quo, which is to be very strict, but rely on the fact that we have `strictNullChecks` turned off in our own code base, so we don't have to worry about this.
2) What @huan is suggesting, namely be liberal about the shape typing and let problems surface during runtime.
3) Be strict, but come up with more specific subtypes for shapes for different layers. For example, an RNN layer can have its specific shape subtype:
`export type RNNInputShape = [null | number, null | number, number]`

and a Conv2D layer can have its input shape subtype:
`export type Conv2DInputShape = [null | number, number, number, number]` (assuming NHWC)

No. 1 is not ideal, because we are just pushing the problem to our users.
No. 3 is the right thing to do, but will involve a lot of work.

No. 2 seems to be an acceptable short-term solution to me. Expecting the type system to catch all runtime issues is unrealistic anyway (consider the fact that some layers don't work correctly if you have a wrong dtype).

Bottom line: I think this PR is good, but we can make a TODO item or feature request issue (low-priority) to add layer-specific input shape types.

huan:
@davidsoergel @caisq Thank you both for the valuable discussion with the help for resolving this issue.

I love TypeScript very much and I use it all the time when I work in JavaScript projects because the typing will help the developers to solve the most problems, however, I do agree with @caisq that the type system can not do every job for us.

I'll add a `TODO item` in my PR, and please let me know if it needs any other changes.

huan:
> Can you add at least a few tests, e.g. building a simple model with an unspecified batch size, and showing that it works for training and inference?

@davidsoergel I have a PR with a working model to demonstrate it at here:

https://github.com/huan/tfjs-examples/blob/de5bb8c7fffd3218c1595f0ada8b58e705c0c576/translation/translation.ts#L207-L210

caisq:
@davidsoergel Any more thoughts on this PR?

nsthorat:
Gentle ping to @davidsoergel to review

rix0rrr:
As far as I can tell, this change requires me to disable `strictNullChecks` for my downstream applications that use TensorflowJS. That's a bit of a bummer because I do enjoy the extra checks on my code that this check gives me. Is there a solution for that?

huan:
@rix0rrr I love `strictNullChecks` too.

Could you please provide a minimum re-producible code(or repo) to show this problem?

PR #498, opened 2019-03-15
Add Noise layer into tfjs

WenheLI:
This PR is about https://github.com/tensorflow/tfjs/issues/1326

I added three noise layers described at Keras documentation
https://keras.io/layers/noise/

Thanks

---
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs-layers/498)

### DIFF
--- a/src/exports_layers.ts
+++ b/src/exports_layers.ts
@@ -672,3 +672,39 @@ export const maxPool1d = maxPooling1d;
 export const maxPool2d = maxPooling2d;
 
 export {Layer, RNN, RNNCell, input /* alias for tf.input */};
+
+/**
+ * @doc {
+ *   heading: 'Layers',
+ *   subheading: 'Noise',
+ *   namespace: 'layers',
+ *   useDocsFrom: 'GaussianNoise'
+ * }
+ */
+export function gaussianNoise(args: GaussianNoiseArgs): GaussianNoise {
+  return new GaussianNoise(args);
+}
+
+/**
+ * @doc {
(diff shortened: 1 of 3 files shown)


WenheLI:
I signed it!

WenheLI:
Thanks @caisq . I will work on the unit test soon. Just one thing I notice that namely most layers are exported by ``export_layers.ts``, do I need to add my layers into this file?

caisq:
@WenheLI Good question! I forgot to comment on that. Please follow the patterns of other layers (e.g., Dense in core.ts) in order to export the factory method of the new layer type under the tf.layers.* namespace.

WenheLI:
Hi, I just added some symbolic tests on the noise layers. Do I need to add more tests to this unit? Since I found it might be hard to test these layers with actual values, I might need some references on it to go further?

WenheLI:
@caisq Thanks for your comments. I have fixed most of your comments. Just one more thing remains to be unfinished that is the unit tests for concrete values. 

Since all the three noise layers will get random results, I am thinking test if their outputs will be a normal distribution that requires a function `jarqueBeraNormalityTest` in `@tensorflow/tfjs-core/src/ops/rand_util` which is not exported. Do I need to manually export it? 
Or I should give these layers a fixed seed and test if the output matches our expected value given the fixed seed?

Thanks!

WenheLI:
@caisq Thanks for your suggestions! I have added the unit test. Let me know if there is any further problem!

PR #499, opened 2019-03-15
Fix l2Normalize rank error in epsilon-flooring of sum

FremyCompany:
See https://stackoverflow.com/questions/55159903/tensorflow-js-error-in-gradient-for-op-maximum-the-gradient-of-input-a-has for details and an example of code that breaks while this bug isn't fixed

---
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs-layers/499)

### DIFF
--- a/src/engine/training_test.ts
+++ b/src/engine/training_test.ts
@@ -496,11 +496,11 @@ describeMathCPUAndGPU('LayersModel.fit', () => {
     expect(history.history.loss.length).toEqual(2);
     expect(history.history.val_loss.length).toEqual(2);
     test_util.expectArraysClose(
-      history.history['loss'] as number[], 
-      [-0.70710688829422, -0.7077317237854004]);
+        history.history['loss'] as number[], 
+        [-0.70710688829422, -0.7077317237854004]);
     test_util.expectArraysClose(
-      history.history['val_loss'] as number[], 
-      [-0.70710688829422, -0.7077317237854004]);
+        history.history['val_loss'] as number[], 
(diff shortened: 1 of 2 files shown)


FremyCompany:
I signed it!

FremyCompany:
Well, I guess I just found another bug... This never happened locally on my model because no vector ever gets so close to 0 that the norm gets to be floored to epsilon. Not sure why I went straight for the integration test instead of a unit test like you proposed, but well, now I guess we might want to address this as well.

Unfortunately, the most obvious fix for this would be to add a derivative function for `onesLike` in tfs-core returning `zeroesLike($a)` and that change is out of scope for this repo. 

I could also change the flooring behavior such that it doesn't require the usage of `onesLike`. That sounds doable, so I'll follow that approach and see if I can get this fixed that way.

FremyCompany:
Here we go, build passed. I took the approach stated above, and replaced the `onesLike` operation with a `fill` function call that doesn't take any tensor input.

FremyCompany:
> We should maybe file a bug on tfjs-core about either getting `onesLike` to remove $x as an input, or return `zeroesLike($x)` as its derivative, but I'll let you take it from there.

FremyCompany:
Thanks for taking a look. Hopefully this should be OK now. 

I'm sorry, I am brand new to Tensorflow, this was actually my first project with it, and I have neither the necessary tooling nor experience to create a reference Python code, but you should have write access to this PR in case this is something you would like to add yourself.

caisq:
Thanks, @FremyCompany !

PR #512, opened 2019-03-30
Update to 1.1.0 and fix unit tests for new API from core.

nsthorat:
Update to 1.1.0 and:

- Add explicit test registration to test_utils. Core no longer automatically registers test backends, each package must do that explicitly.
- All calls to mock window.fetch are now mocking the core util.fetch which is platform-agnostic.
- Calls to get the env flag EPSILON now call directly to the backend().epsilon() (there is no more env flag since it's a backend property)
- Updated unit tests to reflect new changes to optimizers (fewer variables)

Verified we have the same number of tests running before and after locally (4515).

---
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs-layers/512)


nsthorat:
FYI cloud testing is blocked on https://github.com/tensorflow/tfjs-layers/pull/527 but ready to review in the meantime.

PR #513, opened 2019-03-30
Remove getScalar().

nsthorat:
There is no longer any need for this method as we allow passing a number directly to core ops.

Also remove Firefox from local testing (so testing is faster, travis covers ffx).

Fixes https://github.com/tensorflow/tfjs/issues/957

This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs-layers/513)

### DIFF
--- a/karma.conf.js
+++ b/karma.conf.js
@@ -39,7 +39,7 @@ module.exports = function(config) {
     },
     karmaTypescriptConfig,
     reporters: ['progress', 'karma-typescript'],
-    browsers: ['Chrome', 'Firefox'],
+    browsers: ['Chrome'],
     browserStack: {
       username: process.env.BROWSERSTACK_USERNAME,
       accessKey: process.env.BROWSERSTACK_KEY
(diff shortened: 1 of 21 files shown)


nsthorat:
Correct, these will be handled as uniforms uploads to shader programs.

The creation overhead is negligible, uniforms are extremely cheap (no texture allocation, the only overhead allocation if a Tensor object backed by a CPU array). This means you can call x.mul(2) repeatedly with negigible overhead (simply the Tensor object).

With getScalar() you always hold on to memory for new scalars, so if getScalar() is called and there are lots of cache misses we'll be allocating lots of Tensor objects that never get destroyed.

In Node, we create tensor for each of these. There is a trade-off, getScalar() with lots of cache-misses means lots of TFE handles. With this approach we always create and destroy Tensor objects when you pass a non boxed number (e.g. just passing the number to the op). There is an idea to reuse memory internally just like how we do in the webgl texture manager in Node but that's not done yet.

caisq:
@nsthorat Thanks for the detailed reply. Can we wait until I do some detailed profiling of this change in browser and tfjs-node before merging this PR? I just want to be careful.

nsthorat:
Yes, no problem. Let me know!

caisq:
I ran the benchmarks on my local machine multiple times and did not observe any significant changes in the predict() and fit() wall times caused by this change. LGTM.

PR #514, opened 2019-04-02
Fix improper template litteral in topology.ts

FremyCompany:
The current code yields an error message that is less helpful than intended.

---
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs-layers/514)


caisq:
Thanks, @FremyCompany

PR #517, opened 2019-04-03
Add reverseSequences option to RNN

FremyCompany:
Use case: if you try to make your own bidirectional RNN (forwards rnn1, bacwards rnn2, then forwards rnn3 based on output of rnn1+rnn2) then the order returned for a goBackwards sequence is not the same as the one returned otherwise, making the construction of the third rnn difficult.

Simply allowing to reverse the output sequence of a RNN solves that problem.

---
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs-layers/517)

### DIFF
--- a/src/layers/recurrent.ts
+++ b/src/layers/recurrent.ts
@@ -126,6 +126,9 @@ export function standardizeArgs(
  *   expensive concatenation of the stepwise outputs can be omitted unless
  *   the stepwise outputs need to be kept (e.g., for an LSTM layer of which
  *   `returnSequence` is `true`.)
+ * @param reverseStepOutputs If `needsPerStepOutputs` is true, whether the order
+ *   of the outputs per step should be reversed or not (useful if you have
+ *   set `goBackwards` to true and want to perform a bidirectional operation)
  * @returns An Array: `[lastOutput, outputs, newStates]`.
  *   lastOutput: the lastest output of the RNN, of shape `[samples, ...]`.
  *   outputs: tensor with shape `[samples, time, ...]` where each entry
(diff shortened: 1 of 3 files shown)


caisq:
@FremyCompany Thanks for the PR. I'm a worried about the following aspects of the design:
1. The new `reverseSequences` field is not serialized when a RNN layer is saved. So if you save a model with such a layer and load it back, the setting will be lost.
2. There is not equivalent of this new field in either keras-team/keras or tensorflow.keras. So if someone trains and saves a model with an RNN layer with `reverseSequences=true`, convert the model to keras HDF5 format and load it in Python, it won't work as intended.

Note that serialization parity with keras-team/keras and tensorflow.keras is our goal. It may make sense to send a (simultaneous) PR to keras-team/keras and tensorflow/ and see what the reviewers there think. I personally think it's a worthy new feature.

FremyCompany:
@caisq Great feedback, thanks! 

1. I certainly would want this setting to be preserved on serialization. Where would that need to be updated?

2. That sounds like a good idea, but I'm not sure I would have the skills/time to change their implementations to add the feature there. Would opening an issue in their repositories be enough? Or maybe a strawman PR?

FremyCompany:
@caisq The feature was not judged performance-critical enough to warrant an addition to an already complex layer, and I was suggested to use a reverse layer. As far as I know, there is no reverse layer in TensorflowJS; I could make an custom one but then I'd loose serialization. 

Should I open an issue around creating a builtin reverse layer?

caisq:
Closing stale PR. @FremyCompany Feel free to add an issue against tensorflow (https://github.com/tensorflow/tensorflow/issues) to add a reverse layer in Python keras first. Then we can mirror that on the JavaScript side.

PR #538, opened 2019-05-09
Fix metrics examples in comments to make them executable in webpage

syt123450:
MISC

### In this PR

* ~~Fix `useDocsFrom` doesn't work for metrics in tfjs-layers, described in [tensorflow/tfjs#710](https://github.com/tensorflow/tfjs/issues/710).~~
* Fix metrics examples in comments to make them executable in webpage.

### Test

To ensure metrics documentation works well with this fix PR, I built a local website and made a screen shot of the metrics section as shown below:

Screen Shot 2019-05-08 at 2 24 12 PM

### Note

~~In this PR, use `docsForwardAlias` configuration in jsDoc, relative PR about this feature in [tensorflow/tfjs-website#308](https://github.com/tensorflow/tfjs-website/pull/308)~~

---
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs-layers/538)

### DIFF
--- a/src/exports_metrics.ts
+++ b/src/exports_metrics.ts
@@ -16,7 +16,7 @@ import * as metrics from './metrics';
  * @doc {
  *   heading: 'Metrics',
  *   namespace: 'metrics',
- *   useDocsFrom: 'binaryAccuracy'
+ *   useDocsFrom: '_binaryAccuracy'
  * }
  */
 export function binaryAccuracy(yTrue: Tensor, yPred: Tensor): Tensor {
@@ -27,7 +27,7 @@ export function binaryAccuracy(yTrue: Tensor, yPred: Tensor): Tensor {
  * @doc {
  *   heading: 'Metrics',
  *   namespace: 'metrics',
- *   useDocsFrom: 'binaryCrossentropy'
+ *   useDocsFrom: '_binaryCrossentropy'
  * }
  */
 export function binaryCrossentropy(yTrue: Tensor, yPred: Tensor): Tensor {
@@ -38,7 +38,7 @@ export function binaryCrossentropy(yTrue: Tensor, yPred: Tensor): Tensor {
  * @doc {
  *   heading: 'Metrics',
  *   namespace: 'metrics',
- *   useDocsFrom: 'sparseCategoricalAccuracy'
+ *   useDocsFrom: '_sparseCategoricalAccuracy'
  * }
  */
 export function sparseCategoricalAccuracy(
@@ -50,7 +50,7 @@ export function sparseCategoricalAccuracy(
  * @doc {
  *   heading: 'Metrics',
  *   namespace: 'metrics',
- *   useDocsFrom: 'categoricalAccuracy'
+ *   useDocsFrom: '_categoricalAccuracy'
  * }
  */
 export function categoricalAccuracy(yTrue: Tensor, yPred: Tensor): Tensor {
@@ -61,7 +61,7 @@ export function categoricalAccuracy(yTrue: Tensor, yPred: Tensor): Tensor {
  * @doc {
  *   heading: 'Metrics',
  *   namespace: 'metrics',
- *   useDocsFrom: 'categoricalCrossentropy'
+ *   useDocsFrom: '_categoricalCrossentropy'
  * }
  */
 export function categoricalCrossentropy(yTrue: Tensor, yPred: Tensor): Tensor {
@@ -72,7 +72,7 @@ export function categoricalCrossentropy(yTrue: Tensor, yPred: Tensor): Tensor {
  * @doc {
  *   heading: 'Metrics',
  *   namespace: 'metrics',
- *   useDocsFrom: 'precision'
+ *   useDocsFrom: '_precision'
  * }
  */
 export function precision(yTrue: Tensor, yPred: Tensor): Tensor {
@@ -83,7 +83,7 @@ export function precision(yTrue: Tensor, yPred: Tensor): Tensor {
  * @doc {
  *   heading: 'Metrics',
  *   namespace: 'metrics',
- *   useDocsFrom: 'recall'
+ *   useDocsFrom: '_recall'
  * }
  */
 export function recall(yTrue: Tensor, yPred: Tensor): Tensor {
@@ -94,7 +94,7 @@ export function recall(yTrue: Tensor, yPred: Tensor): Tensor {
  * @doc {
  *   heading: 'Metrics',
  *   namespace: 'metrics',
- *   useDocsFrom: 'cosineProximity'
+ *   useDocsFrom: '_cosineProximity'
  * }
  */
 export function cosineProximity(yTrue: Tensor, yPred: Tensor): Tensor {
@@ -105,7 +105,7 @@ export function cosineProximity(yTrue: Tensor, yPred: Tensor): Tensor {
  * @doc {
  *   heading: 'Metrics',
  *   namespace: 'metrics',
- *   useDocsFrom: 'meanAbsoluteError'
+ *   useDocsFrom: '_meanAbsoluteError'
  * }
  */
 export function meanAbsoluteError(yTrue: Tensor, yPred: Tensor): Tensor {
@@ -116,7 +116,7 @@ export function meanAbsoluteError(yTrue: Tensor, yPred: Tensor): Tensor {
  * @doc {
  *   heading: 'Metrics',
  *   namespace: 'metrics',
- *   useDocsFrom: 'meanAbsolutePercentageError'
+ *   useDocsFrom: '_meanAbsolutePercentageError'
  * }
  */
 export function meanAbsolutePercentageError(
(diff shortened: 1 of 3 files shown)


tafsiri:
@syt123450 thanks for the really thorough fix and PRs (including the screenshot)!

@caisq I'm going to cc @nsthorat on this as he was originally on that issue and may have a better sense of whether there is a fundamental reason useDocsFrom doesn't work (it might be the package boundary thing we discussed a while back).

nsthorat:
Instead of this approach can you try adding the namespace to the useDocsFrom (metrics.binaryAccuracy)? If that doesn't work I can fix it with useDocsFrom following namespaces.

syt123450:
Hi, I read all build-script code in [tfjs-website/build-scripts](https://github.com/tensorflow/tfjs-website/tree/master/build-scripts), I think the fundamental reason for the `useDocsFrom` problem is the `method name conflict`. To be more spacific:

I found that when tfjs-website builds `source/_data` from JSDoc, it would save all exported methods into a map `globalSymbolDocMap`([tfjs-website/build-scripts/doc-gen/parser.ts#L170](https://github.com/tensorflow/tfjs-website/blob/master/build-scripts/doc-gen/parser.ts#L170)), the key of the map is the method name .
```js
// Map all symbols to their documentation so we can map useDocFrom aliases.
if (ts.isMethodDeclaration(node) || ts.isFunctionDeclaration(node) ||
    ts.isClassDeclaration(node) || ts.isInterfaceDeclaration(node) ||
    ts.isTypeAliasDeclaration(node)) {
  const symbol = checker.getSymbolAtLocation(node.name);
  const type =
      checker.getTypeOfSymbolAtLocation(symbol, symbol.valueDeclaration!);
  const name = symbol.getName();
  let documentation =
      ts.displayPartsToString(symbol.getDocumentationComment(checker));
  documentation = util.removeStarsFromCommentString(documentation);

  const signature = type.getCallSignatures()[0];

  let params = [];
  if (signature != null && signature.parameters != null) {
    const identifierGenericMap = ts.isMethodDeclaration(node) ?
        util.getIdentifierGenericMap(node) :
        {};

    const isConfigParam = false;
    params = signature.parameters != null ?
        signature.parameters.map(
            symbol => serializeParameter(
                checker, symbol, identifierGenericMap, isConfigParam)) :
        [];
  }
  // ===== HERE, HERE, HERE =====
  // =====Save comments of export assets (for example Method) =====
  // =====The key of the map is the method name =====
  globalSymbolDocMap[name] = {docs: documentation, params};
}
```

And then use `useDocsFrom` to search name in `globalSymbolDocMap` and get the comment information. 

```js
util.replaceUseDocsFromDocStrings(docHeadings, globalSymbolDocMap);
```

syt123450:
So I think the reason for `useDocsFrom` can not work for `metrics` is that tfjs-layers exports metrics in `exports_metrics.ts`, for example, export `binaryAccuracy` [tfjs-layers/src/exports_metrics.ts#L15](https://github.com/tensorflow/tfjs-layers/blob/master/src/exports_metrics.ts#L15)
```js
/**
 * @doc {
 *   heading: 'Metrics',
 *   namespace: 'metrics',
 *   useDocsFrom: 'binaryAccuracy'
 * }
 */
export function binaryAccuracy(yTrue: Tensor, yPred: Tensor): Tensor {
  return metrics.binaryAccuracy(yTrue, yPred);
}
```

The actual comments for `binaryAccuracy` is in `metrics.ts`, [tfjs-layers/src/metrics.ts#L23](https://github.com/tensorflow/tfjs-layers/blob/master/src/metrics.ts#L23)
```js
/**
 * Binary accuracy metric function.
 *
 * `yTrue` and `yPred` can have 0-1 values. Example:
 * ```js
 * const x = tensor2d([[1, 1, 1, 1], [0, 0, 0, 0]], [2, 4]);
 * const y = tensor2d([[1, 0, 1, 0], [0, 0, 0, 1]], [2, 4]);
 * const accuracy = tfl.metrics.binaryAccuracy(x, y);
 * accuracy.print();
 * ```
 *
 * `yTrue` and `yPred` can also have floating-number values between 0 and 1, in
 * which case the values will be thresholded at 0.5 to yield 0-1 values (i.e.,
 * a value >= 0.5 and <= 1.0 is interpreted as 1.
 * )
 * Example:
 * ```js
 * const x = tensor1d([1, 1, 1, 1, 0, 0, 0, 0]);
 * const y = tensor1d([0.2, 0.4, 0.6, 0.8, 0.2, 0.3, 0.4, 0.7]);
 * const accuracy = tf.metrics.binaryAccuracy(x, y);
 * accuracy.print();
 * ```
 *
 * @param yTrue Binary Tensor of truth.
 * @param yPred Binary Tensor of prediction.
 * @return Accuracy Tensor.
 */
export function binaryAccuracy(yTrue: Tensor, yPred: Tensor): Tensor {
  return tidy(() => {
    const threshold = tfc.mul(.5, tfc.onesLike(yPred));
    const yPredThresholded = K.cast(tfc.greater(yPred, threshold), yTrue.dtype);
    return tfc.mean(tfc.equal(yTrue, yPredThresholded), -1);
  });
}
```

When tfjs-website is using build-scripts to generate `globalSymbolDocMap`, as these two method have the same name, they will share the same key `binaryAccuracy` in `globalSymbolDocMap`, so raise conflict.

syt123450:
My basic strategy is to make two methods have the different names in `globalSymbolDocMap`. I added a check before saving comments information into `globalSymbolDocMap`. If `docsForwardAlias` attribute is set, it will use the alias name as the key in `globalSymbolDocMap`. For example, the key of `binaryAccuracy` method in `exports_metrics.ts` is the original method name `binaryAccuracy`, the key of `binaryAccuracy` method in `metrics.ts` is alias name `_binaryAccuracy`.
```js
if (docInfo !== undefined && docInfo !== null && docInfo.docsForwardAlias !== undefined) {
  globalSymbolDocMap[docInfo.docsForwardAlias] = {docs: documentation, params};
} else {
  globalSymbolDocMap[name] = {docs: documentation, params};
}
```

I also noticed that if `@doc` is set for a specific method, this method will be exported to website. But we would not expect the `binaryAccuracy` method in `metrics.ts` to be exported to website, the method to be exported should be `binaryAccuracy` in `export_metrics.ts`. so I added a check for `@doc`:
```js
if (docInfo.heading !== undefined) {
  subheading.symbols.push(docFunction);
}
```

This check makes sure if `heading` attribute are not set for `@doc`, the method will not be exported to website, which ensures that the configuration like `/** @doc { docsForwardAlias: '_binaryAccuracy'} */` will only be used in `useDocsFrom` and would not influence the total number of exported methods.

syt123450:
Other attempts.

I tried different ways to solve this conflict. The most straight forward way is to change the method name in `metrics.ts` and `losses.ts`. For example, change the name of `binaryAccuracy` into `binaryAccuarcyImpl`. However, I found that to change the method name, we may raise a big change:

* Change at least 11 method names in `metrics.ts` and `losses.ts`.
* Change names in test scripts.
* Make the name styles consistent.

syt123450:
At last, I decided to pull request this `docsForwardAlias` way, the advantage of this way is there is no need to change the source code of tfjs-layers. In this PR, I just proposed a solution, there may be more elegant way, I would like to help fix this problem.

Would appreciate hearing your thoughts on this! @nsthorat @tafsiri @caisq

syt123450:
I tried configuration like:
```js
/**
 * @doc {
 *   heading: 'Metrics',
 *   namespace: 'metrics',
 *   useDocsFrom: 'metrics.binaryAccuracy'
 * }
 */
```

I seems that it still not work, I debuged the build-script, the problem is as I said above, the keyn confliction in map.

nsthorat:
I think this block just needs to be aware of the namespace (docInfo.namespace):
https://github.com/tensorflow/tfjs-website/blob/master/build-scripts/doc-gen/util.ts#L144-L160

nsthorat:
Let me know if you want to make that change otherwise I can take it on.

syt123450:
Hi, @nsthorat thanks for the suggestion. Inspired by your suggestion, I use `namespace` to replace previous `docsForwardAlias` attribute. I made `globalSymbolDocMap` aware of the namespace:
```js
let globalMapKey;
if (docInfo != undefined && docInfo != null && docInfo.namespace != undefined) {
  globalMapKey = docInfo.namespace + '.' + name;
} else {
  globalMapKey = name;
}
globalSymbolDocMap[globalMapKey] = {docs: documentation, params};
``` 

Now, the two method of the same name will be saved in `globalSymbolDocMap` would have different keys, if one of them has `namespace` attribute. For example, the `binaryAccuracy` [tfjs-layers/src/exports_metrics.ts#L15](https://github.com/tensorflow/tfjs-layers/blob/master/src/exports_metrics.ts#L15), will have relative key 'metrics.binaryAccuracy'. The `binaryAccuracy` in `metrics.ts` [tfjs-layers/src/metrics.ts#L23](https://github.com/tensorflow/tfjs-layers/blob/master/src/metrics.ts#L23) would have relative key 'binaryAccuracy'.

(I didn't let `replaceUseDocsFromDocStrings` to be aware of `namespace` because I found it to late to do it, as the previous value would be override.)

To make sure the new fix works well, I built a local website and made a new screen shot as shown below:
Screen Shot 2019-05-10 at 6 49 58 PM

nsthorat:
Okay I have a PR that cleans this up, thanks for finding all this. useDocsFrom is completely gone, and we now test on CI: https://github.com/tensorflow/tfjs-layers/pull/541

PR #541, opened 2019-05-14
Add CI testing of runnable snippets.

nsthorat:
Also:
- Remove useDocsFrom. This is an unnecessary layer of indirection. It is much better to have the jsdoc on the method that is actually used (so code editors can see them). This also allows the "source" button in API docs to point to the place where the docs are. In this PR, this allows us to run snippets without following useDocsFrom.
- Fixes all of the snippets that were broken.
- Ignores testing some snippets that use local storage.

---
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs-layers/541)

### DIFF
--- a/cloudbuild.yml
+++ b/cloudbuild.yml
@@ -24,7 +24,7 @@ steps:
   entrypoint: 'yarn'
   args: ['test-snippets']
   id: 'test-snippets'
-  waitFor: ['yarn']
+  waitFor: ['yarn', 'tfjs2keras-py']
 secrets:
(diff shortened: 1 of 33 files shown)


nsthorat:
Blah, ts-node doesn't like .ts files in node_modules so I will publish .js files from core for ./scripts/.

nsthorat:
I need to wait for this PR: https://github.com/tensorflow/tfjs-core/pull/1756

nsthorat:
@caisq any idea what this failure is? https://pantheon.corp.google.com/cloud-build/builds/12c25449-174b-44da-83b5-fe9c8feeb9b8?project=learnjs-174218

PR #544, opened 2019-05-25
Fix onProgress for http url in loadLayersModel

syt123450:
BUG

Issue [tensorflow/tfjs#1610](https://github.com/tensorflow/tfjs/issues/1610) pointed out that `onProgress` callback of tf.loadGraphModel breaks when using HTTP url. While I found that it also breaks in tf.loadLayersModel, for example:

```javascript
// onProgress works
const model = await tf.loadLayersModel('./PATH/TO/MODEL/model.json', {
  onProgress: function(fraction) {
    console.log(fraction);
  }
});
```

```javascript
// onProgress breaks
const model = await tf.loadLayersModel('https://URL/TO/MODEL/model.json', {
  onProgress: function(fraction) {
    console.log(fraction);
  }
});
```
This PR passes `onProgress` to io.getLoadHandlers to fix the bug.

**Note:**
This PR depends on [tensorflow/tfjs-core#1769](https://github.com/tensorflow/tfjs-core/pull/1769)

---
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs-layers/544)

### DIFF
--- a/src/models.ts
+++ b/src/models.ts
@@ -242,7 +242,7 @@ export async function loadLayersModelInternal(
     options = {};
   }
   if (typeof pathOrIOHandler === 'string') {
-    const handlers = io.getLoadHandlers(pathOrIOHandler);
+    const handlers = io.getLoadHandlers(pathOrIOHandler, options.onProgress);
     if (handlers.length === 0) {
       // For backward compatibility: if no load handler can be found,
       // assume it is a relative http path.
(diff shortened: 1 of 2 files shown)


syt123450:
Hi @caisq , thanks for the review. I have added a unit test for it. The new test case break the cloud build, because this PR depends on the fixing in [tensorflow/tfjs-core#1769](https://github.com/tensorflow/tfjs-core/pull/1769). I have used `yarn link-local @tensorflow/tfjs-core` to test in my local environment (with the change in tensorflow/tfjs-core#1769), this test case passed.

caisq:
@syt123450 I see. Let's get the tfjs-core PR reviewed and checked in first. In the meantime, you can add the test, we'll run it again once your fix in tfjs-core is released.

PR #545, opened 2019-05-27
Add LayerNormalization to tfjs

Derek-Wds:
This is about issue [#1548](https://github.com/tensorflow/tfjs/issues/1548).

---
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs-layers/545)


Derek-Wds:
@caisq Hi, thanks for you reply. I will make some changes soon, but I have a question and I hope you could help me with it:
* Since LayerNormalization is doing the same thing as BatchNormalization but on different dimensions, and it uses the function of batch norm. In the unit test of BatchNormalization, it seems that all the tests are based on `functions` instead of BatchNormalization `class`. I wonder if there have other ways to test the LayerNormalization `class`.

Thanks,
Derek

caisq:
@Derek-Wds It's not true that all tests in `normalization_test.ts` are based on functions and not classes. Please look at the test that use `tfl.layers.batchNormalization()`. It may appear to be a function call, but that function actually is a factory function that calls the constructor 'BatchNormalization' under the hood. So it's testing the class.

https://github.com/tensorflow/tfjs-layers/blob/master/src/layers/normalization_test.ts#L301

Derek-Wds:
@caisq Thanks for your response! I'm looking into that.

Derek-Wds:
@caisq Hi, I have added unit test and change the implementation. Can you help me take a look at the error? I cannot run test on my machine (it says something went wrong with the [karma-server]).

Thanks,
Derek

caisq:
@Derek-Wds I'm curious why you closed this PR. Did you run into any trouble? If so, is there anything I can help with? I think LayerNormlization is an important enough new layer type that it should be added.

PR #546, opened 2019-05-27
Fix implementation bug in AlphaDropout and GaussianDropout

syt123450:
BUG

* Fixes [tensorflow/tfjs#1617](https://github.com/tensorflow/tfjs/issues/1617)
* Improves test cases to ensure AlphaDropout and GaussianDropout work well
* Issue mentioned the bug in `AlphaDropout`, while it seems that `GaussianDropout` also has the same bug.

Explanation
* Caused by implementation error, based on `AlphaDropout` and `GaussianDropout` implementation in `Keras` ([GaussianDropout](https://github.com/keras-team/keras/blob/master/keras/layers/noise.py#L91), [AlphaDropout](https://github.com/keras-team/keras/blob/master/keras/layers/noise.py#L162)), should replace `dot` with `mul`.

---
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs-layers/546)

### DIFF
--- a/src/layers/noise.ts
+++ b/src/layers/noise.ts
@@ -139,7 +139,7 @@ export class GaussianDropout extends Layer {
       if (this.rate > 0 && this.rate < 1) {
         const noised = () => {
           const stddev = Math.sqrt(this.rate / (1 - this.rate));
-          return K.dot(input, K.randomNormal(input.shape, 1, stddev));
+          return input.mul(K.randomNormal(input.shape, 1, stddev));
         };
         return K.inTrainPhase(noised, () => input, kwargs['training'] || false);
       }
@@ -238,7 +238,7 @@ export class AlphaDropout extends Layer {
           const b = -a * alphaP * this.rate;
 
           // Apply mask.
-          const x = K.dot(input, keptIdx).add(keptIdx.add(-1).mul(alphaP));
+          const x = input.mul(keptIdx).add(keptIdx.add(-1).mul(alphaP));
 
           return x.mul(a).add(b);
         };
--- a/src/layers/noise_test.ts
+++ b/src/layers/noise_test.ts
@@ -102,7 +102,7 @@ describeMathCPU('GaussianDropout: Symbolic', () => {
 
 describeMathCPUAndGPU('GaussianDropout: Tensor', () => {
   it('GaussianDropout: Predict', () => {
-    const input = ones([2, 2]);
+    const input = ones([1, 2]);
     const gaussianDropoutLayer = tfl.layers.gaussianDropout({rate: 0.5});
     let output = gaussianDropoutLayer.apply(input, {training: false}) as Tensor;
     output = getExactlyOneTensor(output);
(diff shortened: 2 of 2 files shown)


syt123450:
Hi @caisq , thanks for the review.

The reason for `the change from 2 to 1` is that when using the `ones([2, 2])` as input, both `dot` and `mul` can pass the test. However, when using `ones([1, 2])` as input, only `mul` can pass the test, `dot` can not pass the test. In this way, our test cases can make sure `AlphaDropout` and `GaussianDropout` work well.

---

* To explain why both `dot` and `mul` can pass the test cases using `ones([2, 2])` as input.

for a `dot` product `dot(x, y)`(assume x.rank == y.rank == 2), two vector should follow the rule:
```
x.shape[0] == y.shape[1]
x.shape[1] == y.shape[0]
```
The original implementation using `dot`, for example, `GaussianDropout`:
```js
return K.dot(input, K.randomNormal(input.shape, 1, stddev));
```
Assume: `x` is `input`, `y` is `K.randomNormal(input.shape, 1, stddev)`

* When using `ones([2, 2])` as input. The `shape` of x is `[2, 2]`, the `shape` of y is `[2, 2]`.
```
x.shape[0] == x.shape[1] == y.shape[0] == y.shape[1]
```
In this case, although the `dot` implementation is wrong, tests still passed.

* When using `ones([1, 2])` as input. The `shape` of x is `[1, 2]`, the `shape` of y is `[1, 2]`.
```
x.shape[0] != y.shape[1]
x.shape[1] != y.shape[0]
```
In this case, when running the unit tests, it will through exception for the previous `dot` product implementation, while new `mul` implementation still passed.

syt123450:
Hi @caisq , thanks for the review.

The reason for `the change from 2 to 1` is that when using the `ones([2, 2])` as input, both `dot` and `mul` can pass the test. However, when using `ones([1, 2])` as input, only `mul` can pass the test, `dot` can not pass the test. In this way, our test cases can make sure `AlphaDropout` and `GaussianDropout` work well.

---

* To explain why both `dot` and `mul` can pass the test cases using `ones([2, 2])` as input.

For a `dot` product `dot(x, y)`(assume x.rank == y.rank == 2), two vector should follow the rule:
```
x.shape[1] == y.shape[0]
```
The original implementation using `dot`, for example, `GaussianDropout`:
```js
return K.dot(input, K.randomNormal(input.shape, 1, stddev));
```
Assume: `x` is `input`, `y` is `K.randomNormal(input.shape, 1, stddev)`

* When using `ones([2, 2])` as input. The `shape` of x is `[2, 2]`, the `shape` of y is `[2, 2]`.
```
x.shape[0] == x.shape[1] == y.shape[0] == y.shape[1]
```
In this case, although the `dot` implementation is wrong, tests still passed.

* When using `ones([1, 2])` as input. The `shape` of x is `[1, 2]`, the `shape` of y is `[1, 2]`.
```
x.shape[1] != y.shape[0]
```
In this case, when running the unit tests, it will through exception for the previous `dot` product implementation, while new `mul` implementation still passed.

To be more specific, I kept the `dot` implementation, changed the input to be `ones([1, 2])`, and then ran the test cases, I got the following exception.
Screen Shot 2019-05-30 at 9 43 55 PM

caisq:
@syt123450 Got it. Thanks for the detailed explanation.

PR #551, opened 2019-05-31
Enable use of tf.metrics functions in model.fit

syt123450:
FEATURE

This PR enables use of tf.metrics functions in model.fit which was proposed by @caisq and @tafsiri in [tensorflow/tfjs#1171](https://github.com/tensorflow/tfjs/issues/1171).

This PR makes the following function to work.
```js
model.compile({
  optimizer: tf.train.sgd(LEARNING_RATE),
  loss: tf.losses.meanSquaredError,
  metrics: [tf.metrics.meanSquaredError]
});
```
The logs object uses the function name as key, in this case 'meanSquaredError'. I built a local bundle version and tested it with the above case, got the result as the screen shot below:
Screen Shot 2019-05-31 at 3 09 30 AM

In the meantime, this PR extends the acceptant metrics format. Previously, `model.compile` could only accept string map or string array `(string[]|{[outputName: string]: string})`, now it accepts `string|LossOrMetricFn|Array<string|LossOrMetricFn>|{[outputName: string]: string|LossOrMetricFn}`,
for example:
```js
model.compile({
  optimizer: tf.train.sgd(LEARNING_RATE),
  loss: tf.losses.meanSquaredError,
  // string
  metrics: 'mse'
});
```
```js
model.compile({
  optimizer: tf.train.sgd(LEARNING_RATE),
  loss: tf.losses.meanSquaredError,
  // string array
  metrics: ['mse', 'acc']
});
```
```js
model.compile({
  optimizer: tf.train.sgd(LEARNING_RATE),
  loss: tf.losses.meanSquaredError,
  // LossOrMetricFn
  metrics: tf.losses.meanSquaredError
});
```
```js
model.compile({
  optimizer: tf.train.sgd(LEARNING_RATE),
  loss: tf.losses.meanSquaredError,
  // LossOrMetricFn array
  metrics: [tf.losses.meanSquaredError, tf.losses.precision]
});
```
```js
model.compile({
  optimizer: tf.train.sgd(LEARNING_RATE),
  loss: tf.losses.meanSquaredError,
  // Mixed usage
  metrics: [tf.losses.meanSquaredError, 'acc']
});
```
```js
model.compile({
  optimizer: tf.train.sgd(LEARNING_RATE),
  loss: tf.losses.meanSquaredError,
  // dict
  metrics: {'dense_Dense1': tf.losses.meanSquaredError}
});
```

To ensure this PR work well, I
* added unit tests;
* built a bundle file `tf.min.js` locally which can be found [here](https://github.com/syt123450/enable-metrics), and created a [CodePen demo](https://codepen.io/syt123450/pen/eaQKRG), in this CodePen demo, I included my `tf.min.js` bundle file and added the online demos for the above usages.

---
To see the logs from the Cloud Build CI, please join either our [discussion](https://groups.google.com/a/tensorflow.org/forum/#!forum/tfjs) or [announcement](https://groups.google.com/a/tensorflow.org/forum/#!forum/tfjs-announce) mailing list.

---
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs-layers/551)


syt123450:
Hi @caisq , thanks for the review and suggestions. I would look deeper into it and try to find a more elegant way to resolve this problem.

syt123450:
Hi @caisq , inspired by your `reverse lookup` strategy, I adopted another approach to make tf.metrics function wok, could you review the PR again?

As you said, we can have a lookup table to do reverse lookup, I found that we already have two lookup tables `lossesMap` and `metricsMap`. Previously, we only use these tables to do a `one-way` search, which is to search `LossOrMetricsFn` by given `shortcut name`. I update the `getLossOrMetricFnName`, which will use these two tables to search `shortcut name` by given `LossOrMetricsFn`. In this way, we can achieve the `reverse lookup`, and the log information won't be influenced by JS minification.

There might be different ways to implement the lookup table, for example, create a brand new table which contains all `LossOrMetricsFn` and their relative names. While I concerns that with a new lookup table, we need some extra work to maintain the new table information and make the it consist with the previous `lossesMap` and `metricsMap` in the further development, so I didn’t apply this strategy.

Would like to hear you thoughts on this, thanks!

PR #554, opened 2019-06-04
Implements the square and log activations

joyicejin:
implements the square and log activation method in JS
the two activation method in python Code

def square(x):
    return K.square(x)

def log(x):
    return K.log(K.clip(x, min_value=1e-7, max_value=10000))

modified in three files
1. tfjs-layers/src/activation.ts
2. tfjs-layers/src/exports_layers.ts
3. tfjs-layers/src/layers/advanced_activations.ts

---
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs-layers/554)

### DIFF
--- a/readme
+++ b/readme
@@ -1,0 +1,15 @@
+implements the square and log activation method in JS
+the two activation method in python Code
+
+def square(x):
+    return K.square(x)
+
+
+def log(x):
+    return K.log(K.clip(x, min_value=1e-7, max_value=10000))
+
+
+modified in three files
+1. tfjs-layers/src/activation.ts
+2. tfjs-layers/src/exports_layers.ts
+3. tfjs-layers/src/layers/advanced_activations.ts
--- a/src/activations.ts
+++ b/src/activations.ts
@@ -11,6 +11,7 @@
 // Layer activation functions
 import * as tfc from '@tensorflow/tfjs-core';
 import {serialization, Tensor, tidy} from '@tensorflow/tfjs-core';
+import {getScalar} from './backend/state';
 import * as K from './backend/tfjs_backend';
 import {ActivationIdentifier} from './keras_format/activation_config';
 import {deserializeKerasObject} from './utils/generic_utils';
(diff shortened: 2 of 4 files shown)


joyicejin:
I signed it!

davidsoergel:
tfjs-layers implements the Keras API, but neither Keras nor tf.Keras provides these activations (https://keras.io/activations/, https://www.tensorflow.org/api_docs/python/tf/keras/activations).  So, I think we should not include them here either.

joyicejin:
Ok, I know, many thanks for your review

joyicejin:
Sorry for reopen this issue, as I'm a little confused about how to load my customed @tfjs-layers package?

these two funcions are important for me, and I want to load it in my html script file. 
1. Could you give me some instructions for creating my own tf.min.js files( replace the original script src http tf.min.js files)
the replacing code is:
`<script src="../node_modules/@tensorflow/tfjs/dist/tf.min.js"></script>`

2. I'm confused about  how to import these two packages by a script html page of this sentences  in the tfjs-layers code page,
`You can get TensorFlow.js Layers as a module: @tensorflow/tfjs-layers. Note that tfjs-layers has peer dependency on tfjs-core, so if you import @tensorflow/tfjs-layers, you also need to import @tensorflow/tfjs-core.`
Could you tell me the detailed code in HTML script files by the local packages?

davidsoergel:
We do not recommend building customized versions of the TF.js packages themselves.  Instead, you can accomplish what you need using the standard API that we provide.

I see this bit is poorly documented, but anywhere that tfjs-layers takes an activation identifier (a string such as 'relu'), you can instead pass a custom Activation instance.  So you can put your `Square` and `Log` Activation definitions (as you wrote them in `src/activations.ts`) in your own code, and then say things like 

     model.add(tf.layers.dense({ ... , activation: new Square()}));

Note that if you save the resulting model, you'll only be able to reload it in your own code after `serialization.registerClass(Square);` has run (so that the model parser knows how to interpret the nonstandard identifier, 'square').

Re your second question, you are already using the TF.js union package, which includes both tfjs-core and tfjs-layers, so you don't need to worry about importing them separately.

PR #561, opened 2019-06-24
Enable class-weighting fit() and fitDataset()

caisq:
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs-layers/561)


bileschi:
___
*[src/engine/training_weighted_test.ts, line 439 at r1](https://reviewable.io/reviews/tensorflow/tfjs-layers/561#-LiK1ePA5keogXL5THH3:-LiK1ePA5keogXL5THH4:b-vpfais) ([raw file](https://github.com/tensorflow/tfjs-layers/blob/2a835afe034cc5f7ead36ab88a9c046cf461f68d/src/engine/training_weighted_test.ts#L439)):*
> ```
>     // Note that the following values don't match results from Python,
>     // which is a bug in Python TensorFlow / tf.keras. But the final
>     // kernel value does match Python results.
> ```
is there a bug reference number?

PR #574, opened 2019-08-05
Fix input signature for predictOnBatch(); Add unit test

caisq:
BUG

---
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs-layers/574)


caisq:
cc @FremyCompany

PR #578, opened 2019-08-12
Fix memory leak with loaded models containing nested Sequential layers

pvaneck:
If a model loaded with loadLayersModel contains a nested Sequential layer, disposing  of the main model leaves the tensors of the nested model behind. This is because the refCount of the nested model doesn't go to zero during the call to `dispose` since the nested model's refCount is incremented twice: once during sub-model creation and once when the sub-model is added as a layer.

This commit makes it so references to each internally constructed sub-container are tracked and disposed when the parent model is disposed. This ensures the refCount will hit 0.

---
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs-layers/578)

### DIFF
--- a/src/engine/container.ts
+++ b/src/engine/container.ts
@@ -59,6 +59,8 @@ export abstract class Container extends Layer {
   layersByDepth: {[depth: string]: Layer[]};
   nodesByDepth: {[depth: string]: Node[]};
 
+  internalContainerRefs: Container[];
+
   containerNodes = new Set<string>();
 
   // TODO(michaelterry): Add cache support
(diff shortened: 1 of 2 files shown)


pvaneck:
Sure, I can reopen this PR in the new mono-repo when it is ready. 

For the repeated calling issue, seems that with these containers as layers, the first layer.dispose() only decrements the refCount of the container (from 2 to 1), and since the refCount isn't 0, no layers in the container are disposed. With the subsequent call to container.dispose(), the refCount is decremented to 0, and the layers in the container are disposed.

